Issue #171, opened 2026-07-12
Recover from an empty policy file instead of bricking the logic check


SeoyunL:
> **2026-07-16 update.** Two premises below went stale after #155 (PR #179) and the `policy_state` refactor; the core defect stands, but narrower:
>
> - *"the missing-file case ... silently falls back to the default and reports consistent"* — no longer true. Missing + recorded now halts loudly (`MISSING_RECORDED`); missing + never-recorded runs the default **with a warning** (`UNRECORDED_DEFAULT`). The benign-vs-error decision this issue asked for has already been made for the *missing* case.
> - *"There is no way back through the CLI or the web UI"* — `verinote policy reset --force` (`cli.py:138`, `cmd_policy_reset`) now overwrites the file with the shipped default unconditionally, so the KB is no longer permanently bricked.
>
> **What remains open:** an empty or whitespace-only policy file still resolves to `PRESENT` (`resolve_policy` checks only `path.is_file()`, `policy_state.py:113`) and surfaces as `"policy/engine error: program must declare relation/3"` — a cryptic, policy-blaming error that never says "your policy file is empty" and never points at `policy reset --force`. `verinote init` still refuses to re-scaffold (now by design). The remaining ask: give the empty file its own loud, diagnosed state — fold it into the MISSING semantics or report the emptiness with the recovery command, consistent with how #155 treats a lost file. Coordinate with #180 (required `policy_dl`) and #181 (marker fields).

## Symptom

A zero-byte or truncated `policy/logic-policy.dl` **permanently disables the logic check**, and the
scaffold guard blocks its own recovery. There is no way back through the CLI or the web UI.

## Repro

```
policy/logic-policy.dl becomes empty (interrupted write, disk full, user clears it, touch)
-> verify(): ok=False, "policy/engine error: program must declare relation/3"
             (all contradiction checks stop)
-> verinote init: _scaffold_policy() -> None (does not regenerate); file stays empty
```

## Cause

`verinote/pipeline/verify.py:19-24` — `load_policy` returns `""` for an empty file. `""` is not
`None`, so the `DEFAULT_POLICY` fallback never kicks in and the engine rejects the program.

`verinote/cli.py:38-42` and `verinote/web/app.py:84` — `if path.exists(): return None`, so the
scaffold refuses to rewrite the empty file.

This fails closed (no wrong answer), but there is no recovery path short of hand-editing the file.

## Fix direction

[...]

Issue #159, opened 2026-07-12
Stop gitignoring hand-written policy and the review database


SeoyunL:
## Symptom

`.gitignore` classifies irreplaceable, human-authored originals as build output. A routine
`git clean -fdx` deletes the entire KB — every accept/reject decision and the hand-edited policy —
with **zero git history to recover from**.

## Evidence

`.gitignore:11-17` ignores `*.sqlite`, `/data/`, and `*.dl`. But:

- `kb.sqlite` holds **every human accept/reject decision** and the audit log.
- `README.md:33` says ".dl policy/query files remain **editable inputs**" — i.e. human-authored
  source. The `*.dl` pattern is unanchored, so it ignores hand-written policy anywhere in any tree.
- The default KB root is `./data` (`config.py:52-53`) — **inside the repo working tree**.

```
$ git clean -ndx           -> Would remove data/
$ git log --all -- '*.dl'  -> (0 commits)
$ git log --all -- 'data/' -> (0 commits)
```

The generated artifact (`facts/query.dl`) and the human-authored original
(`policy/logic-policy.dl`) are treated identically.

## Fix direction

State in `.gitignore` and README **what is irreplaceable**:

- The generated `facts/query.dl` is an artifact → ignore it specifically.
- `policy/logic-policy.dl` is a source → must not be ignored (at minimum, un-ignore it explicitly).
- Decide and document whether the review DB is meant to be version-controlled; if not, say where
  the user is supposed to back it up, because the review decisions in it cannot be regenerated.

Compounding: when the policy goes missing this way, verification silently uses the default policy
and reports "consistent" (see the load_policy issue).

_Recurrence of a known factlog failure: a `.gitignore` that classified an original as an artifact evaporated 676 approved facts._

SeoyunL:
Fixed in #190.

**The real bug was the unanchored `*.dl` glob**, not the KB being ignored. Extension globs match at any depth, so `*.dl` silently ignored every hand-written policy, test fixture and doc example in the tree — and an uncommitted policy makes `load_policy()` return `None`, verinote falls back to the shipped default policy, and the check still reports **"consistent"**. Green gate, absent rules.

[...]

PR #179, opened 2026-07-12
fix(#155): make a lost logic policy loud instead of silently falling back to DEFAULT_POLICY

### LINKED ISSUE #155: Fail loudly when the KB policy file is missing instead of silently using the default
## Symptom

If `policy/logic-policy.dl` goes missing, verification silently falls back to the shipped
`DEFAULT_POLICY` and reports **`ok=True` — "no findings — knowledge base is consistent."**
The user's own review rules stop being applied, with no warning.

## Repro

KB with `policy/logic-policy.dl` containing `functional("employed_by").` and two conflicting
`employed_by` facts:

- policy present → `ok=False errors=1 ['ERROR functional_conflict: Ada employed_by']`
- policy file deleted → `ok=True errors=0`, body: `no findings — knowledge base is consistent.`

## Cause

`verinote/pipeline/verify.py:19-24`

```python
def load_policy(store) -> str | None:
    path = policy_path(store)
    if path.is_file():
        return path.read_text(encoding="utf-8")
    return None          # caller substitutes DEFAULT_POLICY
```

"No policy file" has two meanings and the code collapses them into the benign one:

- (a) fresh KB, rules were never written → benign, default is fine
- (b) rules existed and the file was lost/moved/renamed → **must be loud**

Aggravated by `.gitignore` excluding `*.dl` (see the gitignore issue), which makes (b) the
likely case in practice.

## Fix direction

Split the two cases explicitly (e.g. a marker/version row recorded when a policy is first
scaffolded, or a policy hash in the store). Only the never-written case may be quiet; a policy
that was present and is now gone must fail the check loudly.

Note: an *empty* policy file already fails loudly (`program must declare relation/3`) — only the
missing-file path is silent.

_Recurrence of a known factlog failure (benign-vs-error conflation; the shape that evaporated 676 approved facts)._

SeoyunL:
## The bug: benign and error were the same code path

`load_policy()` returned `None` whenever `policy/logic-policy.dl` was absent, and every consumer
(`engine/duckdb_backend.py`, `engine/wirelog.py`, and `pipeline/corroboration.py`'s
`store_functional_relations()`) read `None` as "use `DEFAULT_POLICY`". So a KB whose human-written
rules were deleted reported:

```
ok=True, errors=0 — "no findings — knowledge base is consistent."
```

A silently wrong green light. "Policy file absent" has two meanings and code cannot tell them apart:

- (a) new KB, never used rules -> benign
- (b) rules existed and were lost -> error

Since it cannot be *inferred*, the KB now **declares** it.

## The marker (declaration, not inference)

`kb_meta(key, value, updated_at)` in `store/schema.sql` (`init_schema()` runs it on every open, so
`CREATE TABLE IF NOT EXISTS` is also the migration). Key `policy.logic`, value
`{"sha256":..., "recorded_at":..., "origin":"scaffold"|"adopted"|"reset"}`.

- sha256 is **evidence** (the report can say what was there), never a verdict: editing a policy is
  normal, so a **hash mismatch is not an error**.
- The policy body is **never** stored in the DB — the `.dl` file owns it.
- The marker's presence is the only judgement input. No mtime, no git state, no directory
  heuristics. There is also **no `clear_policy_marker()`**: a named way to downgrade a loud state
  back to benign is a bypass waiting to be used.

## Three states, one interpretation point

`pipeline/policy_state.py::resolve_policy(store)`:

| file    | marker  | status               | behaviour                                                |
|---------|---------|----------------------|----------------------------------------------------------|
| present | either  | `PRESENT`            | use the file (unchanged)                                 |
| absent  | present | `MISSING_RECORDED`   | **loud error**: `PolicyMissingError`, engine not run     |
| absent  | absent  | `UNRECORDED_DEFAULT` | `DEFAULT_POLICY` + **explicit warning**; `ok` untouched  |

`load_policy()` delegates here, so the corroboration/acceptance gate became loud without a second
fix. **`PolicyStatus` is the only code that judges policy state** — nothing parses finding strings
to find out (a review round caught `/questions` doing exactly that, see below).

**After this change there is no path that reports "consistent" without saying which policy said so.**

## A halted KB reports; it does not crash, and it takes no new knowledge

> **Wording corrected in round 3.** This section originally said a halted KB "does not accept
> writes", which was stronger than the code: the extraction worker still rewinds its own job
> bookkeeping. The exact, enforced contract is in
> [*What "does not accept writes" actually means*](#what-does-not-accept-writes-actually-means)
> below. Everything in this section about *facts, evidence and policy files* holds as written.

`PolicyMissingError` propagating out of the trust/acceptance path used to 500 `/`, `/review`,
`/workbench` and `/sources` — so the user never reached `/report` and never saw the recovery
instructions. **A 500 is not loud, it is a crash.** Two layers, no per-route `try/except`:

1. **fail-closed middleware** runs before any route. With `MISSING_RECORDED` everything returns
   **409** with `policy_halted.html`, whose text is `policy_missing_message(state)` and nothing else.
   The allowlist is keyed by **(method, path)**, not path alone — a page needed to *diagnose* the
   halt is not licence to *write* under the same prefix:

   | allowed | why |
   |---|---|
   | `GET /report`, `GET /settings`, `GET /static/*` | see the halt and how to recover |
   | `POST /kb/select`, `POST /settings/root` | the only writes that *leave* this KB |

   Everything else fails closed — including `POST /settings/relation-aliases`, because relation
   aliases are a human-written policy that corroboration and query canonicalisation consume, so
   editing them while this KB's rules are not applied is the same category of mistake as accepting a
   fact.
2. **`app.add_exception_handler(PolicyMissingError, ...)`** as a **display-only** backstop: a route
   added outside the guard shows the loud page instead of a stack trace. It deliberately is *not*
   the write barrier — an exception handler runs after the route body, so without the middleware a
   write would commit (SQLite autocommits) and still render a 409: "looks rejected, actually
   written". Write blocking is the middleware's default-deny.

This also closes the write hole: `POST /facts/{id}/accept` used to autocommit the status change and
*then* fail while rendering, so the KB changed while the user saw an error. The guard runs before
the route, so nothing is written. **Blocking accept/reject on a halted KB is the point**: an approval
given while the KB's rules are not applied is a fake human gate. Restoring the policy file unblocks
the KB immediately, and there is deliberately no "re-create the policy" button — recovery is a human
`--force` decision.

## `UNRECORDED_DEFAULT` cannot read as a clean bill of health

This is exactly the state of a KB that lost its policy *before* upgrading, so it matters most for
the original victims of #155. The engine's `"no findings — knowledge base is consistent."` is a claim
about the policy that ran, and this KB's policy did not run — so the sentence is **replaced**, not
merely prefixed with a banner, and the report's OK badge now requires `warnings == 0`.

`ok`/`errors` are deliberately **not** escalated: with no marker there is no evidence a policy was
ever lost, and inventing an error would be inference — the same sin, mirrored. (The CLI does not
print report text, so there is no second surface to keep in sync.)

The sentence itself now has exactly one owner: `engine.NO_FINDINGS_TEXT`, used by both engine
backends and by `verify`'s substitution. It had been spelled out in three places, so re-wording it in
the engine would have silently turned the substitution into a no-op and revived the "consistent"
claim.

## Scaffold guard (without it the fix is pointless)

`cli._scaffold_policy()` and `web._open_root()` used to rewrite `DEFAULT_POLICY` whenever the file
was missing — i.e. they *erased the evidence of the loss*. The condition is now
`file absent AND marker absent`, and the write records `origin="scaffold"`. With a marker and no
file nothing is re-created: `verinote init` exits non-zero with the loud message, and the web app
opens the KB but halts as described above.

## Escape hatch

`verinote policy reset --force` rewrites the default policy and re-records the marker with
`origin="reset"`. Without `--force` it refuses. **No automatic recovery, anywhere.**

## Backwards compatibility (deliberate trade-off)

`ensure_policy_marker(store, root)` runs at every KB-open point (`cli._store()`, `web._open_root()`,
`create_app`):

- file present, no marker -> adopt (`origin="adopted"`); existing KBs behave identically, and any
  *later* deletion is loud
- file present, marker -> refresh the sha256 (never an error)
- file absent, no marker -> nothing (fresh KB -> warning path)
- file absent, marker -> nothing (preserve the loud state)

A KB that *already* had no policy at upgrade time carries no evidence of (a) vs (b), so it gets a
**warning, not a hard failure**. Fatal failure requires evidence — the marker. That is "do not infer",
applied to itself.

## Mutation verification (re-measured on the final tree)

Reverted `resolve_policy`/`verify` to the pre-#155 behaviour (marker check removed, `return None` on
a missing file, no warning annotation) and re-ran `tests/test_policy_state.py`:

```
FAILED tests/test_policy_state.py::test_deleting_a_recorded_policy_is_an_error_not_a_clean_report
FAILED tests/test_policy_state.py::test_resolve_policy_missing_recorded
FAILED tests/test_policy_state.py::test_unrecorded_policy_runs_default_with_a_warning
FAILED tests/test_policy_state.py::test_unrecorded_policy_still_reports_default_policy_errors
FAILED tests/test_policy_state.py::test_web_open_root_does_not_recreate_a_recorded_policy
FAILED tests/test_policy_state.py::test_store_functional_relations_raises_when_policy_is_lost
FAILED tests/test_policy_state.py::test_auto_accept_is_fail_closed_when_the_policy_is_lost
FAILED tests/test_policy_state.py::test_no_get_route_crashes_when_the_policy_is_lost
FAILED tests/test_policy_state.py::test_report_of_unrecorded_policy_kb_is_not_ok
9 failed, 13 passed in 1.59s
```

(This supersedes the "6 red" figure in the first revision of this description. The mutation is scoped
to `verify.py`, which is why the CLI-side guard test `test_init_refuses_to_recreate_a_recorded_policy`
survives it — that guard is pinned separately.) The mutation was reverted; the suite is green.

## Tests

`tests/test_policy_state.py` — 22 tests:

- the issue's measured scenario: `functional("employed_by")` + 2 conflicting facts -> `ok=False`;
  delete the policy file -> **still** `ok=False`, `policy_missing` finding, no "consistent"
- unrecorded default: warning finding + no clean-bill sentence + non-OK badge
- hash mismatch is not an error; `init` records `origin=scaffold`; a pre-marker policy file is adopted
- no re-creation: CLI non-zero exit, web `_open_root` leaves the file absent
- `policy reset --force` (and its refusal without `--force`)
- `store_functional_relations()` and `apply_auto_accept_recommendations()` propagate
  `PolicyMissingError` and change no fact status (fail-closed)
- **every GET route enumerated from `app.routes`** on a halted KB: no 500s, and every non-exempt page
  carries the `policy reset --force` recovery route — so a newly added route is covered automatically
- **every POST/PUT/PATCH/DELETE route enumerated the same way**: 409, no fact status changed, and no
  policy/settings file written behind the halt (`policy/relation-aliases.md`, `config.json`) — this
  locks the default-deny against future routes
- `POST /facts/{id}/accept` on a halted KB: 409 and the fact stays `candidate`
- restoring the policy file unblocks the KB

Full suite: **691 passed, 7 skipped** — no regressions.

## Note for #171 (empty policy file)

Do not "treat whitespace-only as missing". After #155 an empty-but-recorded policy must flow into the
loud path (or a new `EMPTY` member of `PolicyStatus`, which `resolve_policy` is structured to accept).
Re-introducing a silent fallback to `DEFAULT_POLICY` for any state re-opens exactly this bug. Equally:
do not judge policy state by parsing finding strings — findings are human-readable output.

Out of scope, tracked separately: #171 (empty policy), #159 (`.gitignore`), #180 (the engine's own
`None -> DEFAULT_POLICY` fallback), #181 (`recorded_at` immutability).

Closes #155

---

# Review round 2 (@justinjoy): the halt did not actually block writes

The review was right, and the hole was bigger than reported. `_store()` calls `ensure_policy_marker()`,
which for a missing file **only reported the state and returned** — it never blocked. So on a KB with a
deleted policy file, every write command succeeded:

| command | rc (before) | what it wrote into the halted KB |
|---|---|---|
| `ingest input.txt` | **0** | `sources/input.txt` + a `sources` row |
| `seed` | **0** | 4 `facts` rows |
| `query "..."` | **0** | a `questions` row + `facts/query.dl` |

The PR's own contract ("a halted KB must not be written to") was enforced only in the web middleware.

## Fix: one judgement, three enforcement points

**Judgement** is now a single predicate — `policy_state.assert_writable(store)` — raising
`PolicyMissingError` on `MISSING_RECORDED`. Nothing outside `resolve_policy` is allowed to infer policy
state (no ad-hoc `path.exists()` checks), so the enforcement points cannot drift apart in what "halted"
means.

**Enforcement** sits at the three places a write can enter:

1. **CLI — one check in `main()`'s dispatch.** Each subparser declares `set_defaults(halt_safe=...)`;
   `main` reads it with `getattr(args, "halt_safe", False)`, so it is **fail-closed**: a new subcommand
   that forgets to declare one is treated as a write and blocked.
2. **Worker — the write boundary inside `process_extraction_job()`** (`extract.py`): once at job start,
   then **re-checked before every chunk's candidates are persisted**. This is the only way to catch a
   policy deleted *mid-job*: the job is many slow LLM calls and the `Store` was opened before the
   deletion, so no open-time or start-of-command guard can see it. The CLI `sync` path and the web
   worker go through this same function, so both are closed at once.
3. **Web — unchanged behaviour**, but the middleware now calls `assert_writable` instead of
   re-deriving the state.

A Store-layer blanket guard was rejected: writes are spread across many `Store` methods (no single
choke point), and an open-time guard would miss case 2 entirely.

## Command classification

| command | class | on a halted KB |
|---|---|---|
| `init`, `seed`, `sync`, `ingest`, `query`, `repair` | write | **rc=2, writes nothing** |
| `status`, `coverage` | diagnosis | rc=0, still works |
| `policy reset --force` | recovery | rc=0, un-halts the KB |
| `ui` / `serve` | launcher | starts; the web guard blocks writes (409) while `/report` explains the halt |

Diagnosis and recovery stay open on purpose. **Loud is not the same as crash** — a halt the user can
neither inspect nor undo is just a bricked KB, and blocking `ui` would strand them with no route to the
recovery page.

`policy reset --force` runs *on* a halted KB, and works because `write_default_policy()` writes the
**file first and the marker second**: by the time the DB write happens, `resolve_policy` already reads
`PRESENT`. That ordering is now pinned with a comment — reversing those two lines would kill the only
recovery path.

### Where I did not follow the review: `init`

The review suggested exempting `init` from the guard. I did not, and I think exempting it would be a
real bug. `init` on a halted KB re-scaffolds `DEFAULT_POLICY` — which **overwrites the evidence that the
KB's human-written rules were lost with a plausible-looking green KB**, i.e. exactly the failure this PR
exists to prevent. It already returned rc=2 before this round and still does; what's new is that
`init --seed` no longer sneaks demo rows in on its way to the refusal (`_seed()` ran before the policy
check). Recovery is the explicit human gate, `policy reset --force`. Happy to be argued out of this.

(Also noted: there is no `report` **CLI** subcommand — the `/report` **page** is the web one, already
exempt via `_POLICY_GUARD_READ_PATHS`.)

## `delete_meta` (review point 3)

`Store.delete_meta()` had **zero callers** and was `clear_policy_marker()` under another name — sitting
next to the comment explaining why that function deliberately does not exist. It is **deleted**, and
`get_meta`/`set_meta` are demoted to `_get_meta`/`_set_meta`. `kb_meta` has exactly one client (the
policy marker), so no public API can now downgrade a *recorded-but-missing* policy back to
`UNRECORDED_DEFAULT`.

## Tests: 6 new, all mutation-verified

`test_cli_write_commands_refuse_on_halted_kb` (parametrized over ingest/seed/query/sync/repair — asserts
rc≠0, the loss message on stderr, **and** that no `sources/input.txt`, no `facts/query.dl`, and 0
sources/facts/questions rows landed) · `test_init_refuses_on_halted_kb` (rc=2, policy **not**
re-scaffolded, `--seed` writes nothing) · `test_status_and_coverage_survive_halt` (rc=0 — a regression
guard against over-broadening the halt) · `test_policy_reset_force_recovers_halted_kb` (rc=0, file back,
marker `origin="reset"`, and `ingest` works again afterwards) ·
`test_extraction_worker_halts_when_policy_disappears_mid_job` (2-chunk job, client deletes the policy
after chunk 1 → `PolicyMissingError`, and chunk 2's candidate/evidence are **not** in the DB) ·
`test_store_has_no_public_meta_delete`.

**Suite: 701 passed, 7 skipped** (was 691) · `ruff check`: clean.

**Mutation evidence** (each guard removed in turn → the matching test goes red):

| mutation | result |
|---|---|
| remove the CLI dispatch guard | `test_cli_write_commands_refuse_on_halted_kb` — **5 failed** (all params) |
| remove the worker per-chunk re-check | `test_extraction_worker_halts_when_policy_disappears_mid_job` — **failed** (`DID NOT RAISE PolicyMissingError`) |
| widen the guard over `status`/`coverage` | `test_status_and_coverage_survive_halt` — **failed** |

Reproduced the review's sequence by hand afterwards: all six write commands rc=2, `sources/input.txt`
and `facts/query.dl` absent, 0 rows, policy file still gone (loss evidence intact) — while `status`,
`coverage`, `ui` (GET `/report` 200, POST `/review` 409) and `policy reset --force` (rc=0, then `ingest`
rc=0) all work.

**Breaking:** `seed`/`sync`/`ingest`/`query`/`repair` now exit **2** instead of **0** on a KB whose
recorded policy file is missing, and write nothing. `Store.delete_meta()` / `get_meta` / `set_meta` are
no longer public.

---

# Review round 3 (@justinjoy): five holes, closed

Round 2 was measured and confirmed (701 green, mutations red, recovery path alive, fail-closed
dispatch). The follow-up review then found five things still wrong. All five are fixed in `36c987f`.

## 1. `verinote ui` wrote to a halted KB with zero HTTP requests served

`create_app()` ends by calling `_resume_source_extraction_jobs()`, which restarts pending/running jobs
in **worker threads that live outside the HTTP middleware**. The worker's `except Exception` then
caught the `PolicyMissingError` raised by `process_extraction_job` and called
`fail_extraction_job(...)` — a **write**, into the halted KB, from a process that had served no
requests. The PR's claim that "`ui` is only a launcher" was false.

Two layers, both new:

- `_resume_source_extraction_jobs()` is **gated on `assert_writable`** — a halted KB starts no workers.
  Which jobs were resumed is now recorded on `app.state.resumed_job_ids`, so "started no workers" is an
  assertable fact rather than a hope.
- the worker catches `PolicyMissingError` **above `except Exception`** and returns after logging,
  touching no DB. Marking the job `failed` was not just a write, it was a **lie**: the extraction did
  not fail, the KB's rules are gone.

Measured: halted KB + one pending job → `create_app()` → `resumed_job_ids == []` and the **db file is
byte-for-byte identical**.

## 2. A policy lost mid-`sync` was a traceback, not a loud failure

`cmd_sync` caught only `LLMError`, so `PolicyMissingError` reached `main()` and printed a stack trace —
the PR breaking its own "loud is not a crash" rule, and burying the message that names both recovery
routes.

Handled in **`main()`**, once, for the same reason the refusal already lives there ("a guard sprinkled
per-command is a guard the next command will forget"): `PolicyMissingError` → `error: <message>` on
stderr, **rc=2**. This covers `sync`, `repair`, `query` and any future command. A per-command handler in
`cmd_sync` would have been redundant with it — and a redundant guard cannot be mutation-tested, so it is
not there.

## 3. The job-start `assert_writable` had no test (it survived deletion)

Correct — the reviewer's mutation passed. Now pinned by
`test_process_extraction_job_on_halted_kb_writes_nothing_and_calls_no_llm`: on a halted KB the fake
client is called **0 times** and the job row, chunk rows and `runs` table are unchanged (`get_run(1) is
None`).

## 4. What "does not accept writes" actually means

The reviewer is right that the old wording was stronger than the code, and picking option (a) *or* (b)
was a false choice — (a) alone leaves a zombie job. **The contract, stated honestly and now enforced and
tested:**

> A halted KB gains **no candidate fact and no evidence row — ever**. Job/chunk bookkeeping is not
> equally absolute, because the policy can vanish *during* the LLM call of a chunk that was already
> claimed. So bookkeeping is only ever **rewound, never advanced**.

Concretely, in `process_extraction_job`:

- `assert_writable` now runs **before `mark_chunk_running`**, not after. No chunk after the deletion is
  claimed (`status='running'`, `attempts+1`) for work that will never happen.
- the one chunk in flight when the file vanished is **rolled back to `pending`**, which drops the job
  back to `pending` — resumable once a human runs `policy reset --force`. **No zombie `running` job.**
- the `runs` row (opened while the KB was still healthy) gets an honest summary instead of staying
  blank: `"halted: this KB's recorded logic policy file is missing; the job was rolled back to pending
  and no candidate facts were written"`.

The one residue is the in-flight chunk's `attempts` counter, which is not rewound — it truthfully
records that the chunk *was* attempted. Final state is pinned by tests in both directions
(`test_halt_mid_job_rewinds_the_job_to_pending`, `test_halt_between_chunks_never_claims_the_next_chunk`).

## 5. The CLI's only diagnostic surface stayed silent about the halt — this was #155 itself

The sharpest of the five. `verinote status` on a halted KB printed a tidy summary and exited 0, with not
one word about the missing policy; so did `coverage`. There is no CLI `verify`/`report`, so for anyone
not using the web UI, the only way to learn the KB was halted was to try a write and get refused. **"A KB
whose rules are gone must not look fine" was still true on the CLI.**

Every `halt_safe=True` command now reads `resolve_policy` and prints a banner on **stderr**:

- `MISSING_RECORDED` → the full `policy_missing_message` (names both recovery routes)
- `UNRECORDED_DEFAULT` → `POLICY_UNRECORDED_BANNER`

**Exit codes are unchanged** — diagnosis stays usable and pipeable under a halt; the banner is on stderr
so stdout stays parseable. Judgement is `resolve_policy` only; the file is never inspected directly.
A healthy KB prints nothing (`test_status_on_a_healthy_kb_says_nothing_about_the_policy`) — a banner that
is always on is wallpaper, not a signal.

## Verification

**Suite: 711 passed, 7 skipped** (was 701) · `ruff check`: clean · **10 new tests**.

Mutation evidence — each new guard deleted in turn, the matching test goes red:

| # | guard removed | test | result |
|---|---|---|---|
| 1a | `assert_writable` in `_resume_source_extraction_jobs` | `test_create_app_on_halted_kb_resumes_no_jobs_and_touches_no_row` | **RED** |
| 1b | worker's `except PolicyMissingError` (falls through to `except Exception`) | `test_web_worker_never_marks_a_halted_kb_job_failed` | **RED** |
| 2 | `main()`'s `PolicyMissingError` backstop | `test_sync_reports_a_policy_lost_mid_run_instead_of_crashing` | **RED** |
| 3 | job-start `assert_writable` | `test_process_extraction_job_on_halted_kb_writes_nothing_and_calls_no_llm` | **RED** |
| 4a | pre-claim `assert_writable` in `_process_chunks` | `test_halt_between_chunks_never_claims_the_next_chunk` | **RED** |
| 4b | `_rollback_halted_job` | `test_halt_mid_job_rewinds_the_job_to_pending` | **RED** |
| 5 | `_warn_on_degraded_policy` | `test_status_and_coverage_announce_the_halt_on_stderr`, `test_status_announces_an_unrecorded_policy` | **RED** |

Run by hand on a real KB, in order:

| check | result |
|---|---|
| halted KB + pending job → `create_app()` | `resumed_job_ids == []`, **db bytes identical** |
| halted KB → policy deleted mid-`sync` | **rc=2**, loud message, **no traceback**; job back to `pending`, 0 facts, run summary = the halt line |
| halted KB → `status` / `coverage` | **rc=0**, full output on stdout, halt banner on stderr |
| `policy reset --force` → `ingest` | rc=0, then rc=0 — the KB accepts writes again |

## Merge note (conflicts with #186)

#186 (`issue-165-init-seed-named-kb`) touches `verinote/cli.py` in three of the same places
(`cmd_init`'s imports, the `set_defaults` lines for `init`/`seed`, and `main()`'s `cfg = ...` vs its
`_config_for(args)`). **Taking only #186's side of those hunks silently drops the `halt_safe` declarations
and the guard.** When resolving:

- every subparser must keep `halt_safe=...` in its `set_defaults`;
- in `main()`, the `_refuse_on_halted_kb` / `_warn_on_degraded_policy` / `PolicyMissingError` block must
  stay, **after** `_config_for(args)`.

`init`/`seed` losing it would fail closed (annoying, safe); `status`/`coverage`/`ui`/`policy reset`
losing it would **brick the recovery path**.

## Breaking (updated)

In addition to round 2's list: `verinote status` / `coverage` (and the other `halt_safe` commands) now
write a warning banner to **stderr** on a KB whose recorded policy is missing or was never recorded.
Exit codes and stdout are unchanged. An extraction job interrupted by a halt is now left `pending`
(resumable), where it would previously have been left `running` (web) or `failed` (web `except
Exception`).

### DIFF
--- a/tests/test_policy_state.py
+++ b/tests/test_policy_state.py
@@ -521,3 +521,163 @@ def test_report_of_unrecorded_policy_kb_is_not_ok(tmp_path, monkeypatch):
     assert "WARNINGS" in resp.text
     assert "policy_unrecorded" in resp.text
     assert "knowledge base is consistent" not in resp.text
+
+
+# --- a halted KB takes no writes: CLI, worker, and the recovery paths that must
+# --- keep working (the contract this PR claimed but did not enforce)
+
+
+def _halted_cli_kb(tmp_path, monkeypatch):
+    """A CLI-visible KB whose scaffolded policy file has been deleted."""
+    _env(monkeypatch, tmp_path)
+    assert cli.main(["init"]) == 0
+    policy = tmp_path / POLICY_RELPATH
+    policy.unlink()
+    return policy
+
+
+def _kb_rows(tmp_path):
+    store = Store(tmp_path / "kb.sqlite")
+    store.init_schema()
+    try:
+        return (len(store.sources()), len(store.facts()), len(store.questions()))
+    finally:
+        store.close()
+
+
+@pytest.mark.parametrize(
+    "argv",
+    [
+        pytest.param(["ingest", "<src>"], id="ingest"),
+        pytest.param(["seed"], id="seed"),
+        pytest.param(["query", "who is Ada?"], id="query"),
+        pytest.param(["sync", "<src>"], id="sync"),
+        pytest.param(["repair"], id="repair"),
+    ],
+)
+def test_cli_write_commands_refuse_on_halted_kb(tmp_path, monkeypatch, capsys, argv):
+    """Every write command halts — and writes *nothing* — while the policy is lost."""
+    policy = _halted_cli_kb(tmp_path, monkeypatch)
+    src = tmp_path / "input.txt"
+    src.write_text("Ada is_a engineer\n", encoding="utf-8")
+
+    rc = cli.main([a.replace("<src>", str(src)) for a in argv])
+
+    assert rc != 0
+    err = capsys.readouterr().err
+    assert "policy file" in err and "missing" in err
+    # nothing was written: no registered source file, no rows, no query file
+    assert not (tmp_path / "sources" / "input.txt").exists()
+    assert not (tmp_path / "facts" / "query.dl").exists()
+    assert _kb_rows(tmp_path) == (0, 0, 0)
+    assert not policy.exists()  # the evidence of the loss is intact
+
+
+def test_init_refuses_on_halted_kb(tmp_path, monkeypatch, capsys):
+    """`init` is not exempt: re-scaffolding would paper over the lost policy.
+
+    Rewriting the default policy here would turn a KB whose human-written rules
+    are gone into a plausible-looking green KB. Recovery is an explicit human act
+    (`policy reset --force`), so `init` stays a refused write — and `--seed` must
+    not slip demo rows into the halted KB on its way to the refusal either.
+    """
+    policy = _halted_cli_kb(tmp_path, monkeypatch)
+
+    assert cli.main(["init", "--seed"]) == 2
+
+    assert "policy file" in capsys.readouterr().err
+    assert not policy.exists()
+    assert _kb_rows(tmp_path) == (0, 0, 0)
+
+
+def test_status_and_coverage_survive_halt(tmp_path, monkeypatch, capsys):
+    """Read-only diagnosis keeps working — a halt you cannot inspect is a brick."""
+    _halted_cli_kb(tmp_path, monkeypatch)
+
+    assert cli.main(["status"]) == 0
+    assert cli.main(["coverage"]) == 0
+
+    out = capsys.readouterr().out
+    assert "KB:" in out
+    assert "coverage:" in out
+
+
+def test_policy_reset_force_recovers_halted_kb(tmp_path, monkeypatch, capsys):
+    """The recovery path runs *on* a halted KB, and really un-halts it."""
+    policy = _halted_cli_kb(tmp_path, monkeypatch)
+    src = tmp_path / "input.txt"
+    src.write_text("Ada is_a engineer\n", encoding="utf-8")
+    assert cli.main(["ingest", str(src)]) != 0  # halted before recovery
+
+    assert cli.main(["policy", "reset", "--force"]) == 0
+
+    assert policy.read_text(encoding="utf-8") == DEFAULT_POLICY
+    store = Store(tmp_path / "kb.sqlite")
+    store.init_schema()
+    marker = store.policy_marker()
+    store.close()
+    assert marker is not None and marker["origin"] == "reset"
+    # and the KB accepts writes again — recovery that leaves it halted is no recovery
+    capsys.readouterr()
+    assert cli.main(["ingest", str(src)]) == 0
+    assert (tmp_path / "sources" / "input.txt").is_file()
+
+
+class _PolicyDeletingClient:
+    """Deletes the KB's policy file just as the *second* chunk is extracted."""
+
+    name = "fake"
+
+    def __init__(self, policy_path):
+        self.policy_path = policy_path
+        self.calls = 0
+
+    def extract_facts(self, *, source_text: str, schema_hint: str = ""):
+        from verinote.llm.base import ExtractedFact
+
+        self.calls += 1
+        if self.calls == 2:
+            self.policy_path.unlink()
+        return [ExtractedFact(source_text, "is_a", "chunk", 0.9)]
+
+
+def test_extraction_worker_halts_when_policy_disappears_mid_job(tmp_path):
+    """A job started on a healthy KB must stop the moment the policy vanishes.
+
+    The CLI's start-of-command check cannot see this: the job was already running
+    (and the Store already open) when the file was deleted. Only a re-check at the
+    per-chunk write boundary keeps chunk 2 out of a KB with no rules.
+    """
+    from verinote.pipeline.extract import process_extraction_job
+
+    store = _store(tmp_path)
+    path = _write_policy(tmp_path)
+    store.record_policy_marker(policy_sha256(path.read_text(encoding="utf-8")), origin="scaffold")
+    source_id = store.add_source("sources/a.txt")
+    job_id = store.create_extraction_job(
+        source_id=source_id, provider="fake", model="m", total_chunks=2
+    )
+    store.add_source_chunks(job_id=job_id, source_id=source_id, chunks=["alpha", "beta"])
+
+    client = _PolicyDeletingClient(path)
+    with pytest.raises(PolicyMissingError):
+        process_extraction_job(store, client, job_id=job_id)
+
+    assert client.calls == 2  # the second chunk was extracted but never persisted
+    facts = store.facts()
+    assert [f["subject"] for f in facts] == ["alpha"]
+    snippets = [e["snippet"] for f in facts for e in store.fact_evidence(f["id"])]
+    assert "beta" not in snippets
+    store.close()
+
+
+def test_store_has_no_public_meta_delete(tmp_path):
+    """No public API may drop a policy marker — that is the silent-fallback bug."""
+    store = _store(tmp_path)
+    public = {name for name in dir(store) if not name.startswith("_")}
+    store.close()
+
+    assert "delete_meta" not in public
+    assert "set_meta" not in public
+    assert "get_meta" not in public
+    assert "clear_policy_marker" not in public
--- a/tests/test_policy_state.py
+++ b/tests/test_policy_state.py
@@ -1,0 +1,292 @@
+# SPDX-License-Identifier: MPL-2.0
+"""A lost logic policy must never read as a consistent KB (#155)."""
+
+import pytest
+
+import verinote.cli as cli
+from verinote.engine import DEFAULT_POLICY
+from verinote.pipeline.corroboration import store_functional_relations
+from verinote.pipeline.policy_state import (
+    POLICY_RELPATH,
+    PolicyMissingError,
+    PolicyStatus,
+    policy_sha256,
+    resolve_policy,
+)
+from verinote.pipeline.verify import load_policy, verify
+from verinote.store import Store
+
+# A human-written policy: `employed_by` is single-valued for a subject. This is
+# exactly the rule that evaporates in #155 when the policy file is deleted.
+FUNCTIONAL_POLICY = (
+    ".decl relation(subject: symbol, rel: symbol, object: symbol)\n"
+    ".decl functional(rel: symbol)\n"
+    ".decl error_functional_conflict(subject: symbol, rel: symbol)\n"
+    'functional("employed_by").\n'
+    "error_functional_conflict(S, R) :-\n"
+    "    relation(S, R, A), relation(S, R, B), functional(R), A != B.\n"
+)
+
+
+def _store(tmp_path) -> Store:
+    store = Store(tmp_path / "kb.sqlite")
+    store.init_schema()
+    return store
+
+
+def _write_policy(tmp_path, text: str = FUNCTIONAL_POLICY):
+    path = tmp_path / POLICY_RELPATH
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(text, encoding="utf-8")
+    return path
+
+
+def _conflicting_facts(store: Store) -> None:
+    source = store.add_source("sources/a.txt")
+    store.add_fact("Ada", "employed_by", "AcmeCorp", status="confirmed", source_id=source)
+    store.add_fact("Ada", "employed_by", "BetaCorp", status="confirmed", source_id=source)
+
+
+def _env(monkeypatch, tmp_path):
+    monkeypatch.setenv("VERINOTE_ROOT", str(tmp_path))
+    monkeypatch.setenv("VERINOTE_PROVIDER", "anthropic")
+
+
+# --- 1. the reported bug: deleting a recorded policy used to turn the KB green ---
+
+
+def test_recorded_policy_catches_conflict(tmp_path):
+    store = _store(tmp_path)
+    path = _write_policy(tmp_path)
+    store.record_policy_marker(policy_sha256(path.read_text(encoding="utf-8")), origin="scaffold")
+    _conflicting_facts(store)
+
+    rep = verify(store)
+    assert rep.ok is False
+    assert rep.errors == 1
+    assert "functional_conflict" in "\n".join(rep.findings)
+
+
+def test_deleting_a_recorded_policy_is_an_error_not_a_clean_report(tmp_path):
+    store = _store(tmp_path)
+    path = _write_policy(tmp_path)
+    store.record_policy_marker(policy_sha256(path.read_text(encoding="utf-8")), origin="scaffold")
+    _conflicting_facts(store)
+    path.unlink()
+
+    rep = verify(store)
+    assert rep.ok is False
+    assert rep.errors == 1
+    assert any("policy_missing" in finding for finding in rep.findings)
+    assert "consistent" not in rep.text
+    # both recovery routes are named, and neither of them is automatic
+    assert "policy reset --force" in rep.text
+    assert "version control" in rep.text
+
+
+def test_resolve_policy_missing_recorded(tmp_path):
+    store = _store(tmp_path)
+    path = _write_policy(tmp_path)
+    store.record_policy_marker(policy_sha256(path.read_text(encoding="utf-8")), origin="scaffold")
+    path.unlink()
+
+    state = resolve_policy(store)
+    assert state.status is PolicyStatus.MISSING_RECORDED
+    assert state.marker is not None
+    with pytest.raises(PolicyMissingError):
+        load_policy(store)
+
+
+def test_hash_mismatch_is_not_an_error(tmp_path):
+    """Editing a policy is normal: the sha256 is evidence, never a verdict."""
+    store = _store(tmp_path)
+    _write_policy(tmp_path)
+    store.record_policy_marker("0" * 64, origin="scaffold")
+
+    state = resolve_policy(store)
+    assert state.status is PolicyStatus.PRESENT
+    assert verify(store).ok is True
+
+
+# --- 2. fresh KB: no file, no marker -> default policy, but loudly ---
+
+
+def test_unrecorded_policy_runs_default_with_a_warning(tmp_path):
+    store = _store(tmp_path)
+    store.add_fact("Ada", "is_a", "engineer", status="confirmed")
+
+    rep = verify(store)
+    assert rep.ok is True
+    assert rep.errors == 0
+    assert rep.warnings >= 1
+    assert any("policy_unrecorded" in finding for finding in rep.findings)
+    assert "shipped default" in rep.text
+    assert resolve_policy(store).status is PolicyStatus.UNRECORDED_DEFAULT
+    assert load_policy(store) is None
+
+
+def test_unrecorded_policy_still_reports_default_policy_errors(tmp_path):
+    store = _store(tmp_path)
+    store.add_fact("Org", "established_on", "2020", status="confirmed")
+    store.add_fact("Org", "established_on", "2021", status="confirmed")
+
+    rep = verify(store)
+    assert rep.ok is False
+    assert rep.errors > 0
+    assert any("policy_unrecorded" in finding for finding in rep.findings)
+
+
+# --- 3. init scaffolds a policy file and records the marker ---
+
+
+def test_init_records_a_scaffold_marker(tmp_path, monkeypatch):
+    _env(monkeypatch, tmp_path)
+    assert cli.main(["init"]) == 0
+
+    path = tmp_path / POLICY_RELPATH
+    assert path.is_file()
+    store = Store(tmp_path / "kb.sqlite")
+    store.init_schema()
+    marker = store.policy_marker()
+    assert marker is not None
+    assert marker["origin"] == "scaffold"
+    assert marker["sha256"] == policy_sha256(DEFAULT_POLICY)
+    store.close()
+
+
+# --- 4. backwards compatibility: a pre-marker KB adopts its policy on open ---
+
+
+def test_existing_policy_file_is_adopted_on_open(tmp_path, monkeypatch):
+    _env(monkeypatch, tmp_path)
+    path = _write_policy(tmp_path)
+    store = Store(tmp_path / "kb.sqlite")
+    store.init_schema()
+    assert store.policy_marker() is None  # pre-#155 KB
+    store.close()
+
+    assert cli.main(["status"]) == 0
+
+    store = Store(tmp_path / "kb.sqlite")
+    store.init_schema()
+    marker = store.policy_marker()
+    assert marker is not None
+    assert marker["origin"] == "adopted"
+    assert marker["sha256"] == policy_sha256(path.read_text(encoding="utf-8"))
+    store.close()
+
+
+def test_kb_with_no_policy_file_is_not_adopted(tmp_path, monkeypatch):
+    """An upgraded KB that already had no policy has no evidence either way."""
+    _env(monkeypatch, tmp_path)
+    store = Store(tmp_path / "kb.sqlite")
+    store.init_schema()
+    store.close()
+
+    assert cli.main(["status"]) == 0
+
+    store = Store(tmp_path / "kb.sqlite")
+    store.init_schema()
+    assert store.policy_marker() is None
+    assert resolve_policy(store).status is PolicyStatus.UNRECORDED_DEFAULT
+    store.close()
+
+
+# --- 5. a recorded-but-missing policy is never silently re-created ---
+
+
+def test_init_refuses_to_recreate_a_recorded_policy(tmp_path, monkeypatch, capsys):
+    _env(monkeypatch, tmp_path)
+    assert cli.main(["init"]) == 0
+    path = tmp_path / POLICY_RELPATH
+    path.unlink()
+
+    rc = cli.main(["init"])
+    assert rc != 0
+    assert not path.exists()  # re-creating it would hide the loss
+    err = capsys.readouterr().err
+    assert "policy reset --force" in err
+
+
+def test_web_open_root_does_not_recreate_a_recorded_policy(tmp_path, monkeypatch):
+    from fastapi.testclient import TestClient
+
+    from verinote.config import Config
+    from verinote.web.app import create_app
+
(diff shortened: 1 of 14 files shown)


justinjoy (changes requested):
리뷰 결과, 현재 상태로는 변경 요청이 필요합니다.

1. **CLI write path가 halted KB를 계속 변경합니다.**

웹 middleware는 recorded-but-missing policy 상태에서 대부분의 쓰기를 409로 막지만, CLI 경로는 같은 fail-closed 정책을 적용하지 않습니다. `_store()`는 marker adopt/refresh만 수행하고 halt 상태를 차단하지 않기 때문에 `verinote ingest`는 `ingest_file()`로 source 파일과 DB row를 생성할 수 있고, `verinote query`는 question 추가 및 `facts/query.dl` 갱신까지 진행할 수 있습니다.

로컬에서 `init -> policy/logic-policy.dl 삭제 -> verinote ingest`를 재현했을 때 `rc=0`으로 `sources/input.txt`와 `sources` DB row가 생성되는 것을 확인했습니다. PR 본문의 “halted KB reports; it does not crash, and it does not accept writes” 계약과 충돌합니다. CLI의 mutating command에도 동일한 policy halt guard가 필요합니다. 최소한 `init`/`status`/`report`/`policy reset --force` 같은 복구·진단 경로를 제외한 write command는 `PolicyStatus.MISSING_RECORDED`에서 실패해야 합니다.

2. **백그라운드 extraction worker가 middleware 밖에서 계속 쓸 수 있습니다.**

`_start_source_extraction()`은 HTTP middleware 이후 별도 thread에서 독립 `Store`를 열고 `process_extraction_job()`을 호출합니다. 업로드/분석 job이 시작된 뒤 policy 파일이 삭제되면, worker는 halt 상태를 다시 확인하지 않고 job/chunk/run 상태와 candidate facts/evidence를 계속 기록할 수 있습니다. 신규 HTTP 요청은 막히지만 이미 시작된 writer는 막히지 않는 구조입니다. worker 시작 직후와 candidate 쓰기 전 등 적어도 write boundary에서 policy state를 다시 확인하거나, 공통 write guard를 두는 편이 필요합니다.

3. **`delete_meta()`가 policy marker 삭제 우회를 제공합니다.**

`Store`에는 의도적으로 `clear_policy_marker()`가 없다고 설명하지만, public `delete_meta(key)`가 추가되어 `delete_meta(POLICY_MARKER_KEY)`로 같은 효과를 낼 수 있습니다. 현재 호출자는 없지만, 이 API가 남아 있으면 이후 코드가 실수로 recorded-but-missing 상태를 `UNRECORDED_DEFAULT`로 다운그레이드할 수 있습니다. `policy.logic` 삭제를 거부하거나 generic metadata delete를 private/internal로 제한하는 것이 안전합니다.

검증 참고:

- GitHub CI는 `ruff`, `pytest (py3.11)`, `pytest (py3.12)`, `pytest (py3.13)` 모두 통과했습니다.
- 로컬에서도 `tests/test_policy_state.py`는 `22 passed`, 전체 suite는 `691 passed, 7 skipped`였습니다.
- 위 CLI write-hole은 추가 수동 재현으로 확인했습니다.

SeoyunL:
@justinjoy thanks — you were right, and the hole was wider than the report. Pushed db0c1f8.

**What was broken.** `_store()` → `ensure_policy_marker()` only *reported* the missing-policy state and returned; it never blocked. So on a halted KB `ingest`, `seed`, `query` (and `sync`, `repair`) all returned **rc=0** and really wrote — `sources/input.txt` + a `sources` row, 4 `facts` rows, a `questions` row + `facts/query.dl`. The PR's own "a halted KB takes no writes" contract was only ever enforced in the web middleware.

**What I changed.** One judgement, three enforcement points:

- `policy_state.assert_writable(store)` is now the single predicate; nothing outside `resolve_policy` infers policy state.
- **CLI:** one check in `main()`'s dispatch. Subparsers declare `set_defaults(halt_safe=...)` and it's read **fail-closed** (`getattr(args, "halt_safe", False)`), so a future subcommand that forgets to declare one is blocked by default rather than silently exempt.
- **Worker:** `process_extraction_job()` re-checks at the **per-chunk write boundary**, not just at job start. A job is many slow LLM calls with the `Store` already open, so a policy deleted mid-job is invisible to any start-of-command or open-time guard — this was the one case a Store-layer guard could not have caught, which is why I didn't go that route.
- **Web:** behaviour unchanged, just switched onto the same predicate so the three can't drift.

Diagnosis and recovery stay open: `status`, `coverage`, `policy reset --force`, and `ui` (which starts, serves `/report` at 200, and 409s writes). Blocking those would make the halt a brick rather than a warning.

**`delete_meta` (your point 3):** deleted. Zero callers, and it was `clear_policy_marker()` under another name — sitting right next to the comment explaining why that function deliberately doesn't exist. `get_meta`/`set_meta` are now `_get_meta`/`_set_meta`, so no public API can downgrade a recorded-but-missing policy back to a benign default.

**Where I didn't follow you: `init`.** You suggested exempting it from the guard. I kept it refused (rc=2, as it already was), because `init` on a halted KB would re-scaffold `DEFAULT_POLICY` — overwriting the *evidence that the KB's human-written rules were lost* with a plausible-looking green KB, which is the precise failure this PR exists to prevent. What did change: `init --seed` no longer sneaks demo rows in on its way to the refusal (`_seed()` ran before the policy check). Recovery is the explicit human gate, `policy reset --force` — which works on a halted KB because `write_default_policy()` writes the file *before* the marker (now pinned with a comment; reversing it would kill the only recovery path). If you still want `init` exempt, say so and I'll flip it — but I'd rather not.

Small note: there's no `report` **CLI** subcommand; the `/report` **page** is the web one and was already exempt via `_POLICY_GUARD_READ_PATHS`.

**Verification.** 6 new tests (write commands parametrized over ingest/seed/query/sync/repair, asserting rc≠0 *and* zero rows / no files written; init; status+coverage survive; reset recovers and `ingest` works after; the mid-job worker case; no public meta-delete). Suite **701 passed, 7 skipped** (was 691), `ruff check` clean. Each guard removed in turn goes red — CLI guard → 5 failures; worker per-chunk re-check → `DID NOT RAISE PolicyMissingError`; widening the guard over `status`/`coverage` → that test fails. I also re-ran your exact repro by hand: all writes rc=2 with nothing written, policy file still gone.

justinjoy (approved):
Re-checked the updated PR. The previously requested CLI write guard, background extraction write boundary, and public metadata deletion issues are addressed. Local and GitHub checks are green.

SeoyunL:
@justinjoy Round-2 review: all five are fixed in `36c987f`. The two that were genuinely embarrassing:

**(1) `verinote ui` wrote to a halted KB with zero requests served.** You're right, and the PR body's "`ui` is just a launcher" was false. `create_app()` → `_resume_source_extraction_jobs()` → worker threads *outside* the middleware → `except Exception` → `fail_extraction_job`. Fixed at both ends: the resume path is gated on `assert_writable` (and now records `app.state.resumed_job_ids`, so "started no workers" is assertable), and the worker catches `PolicyMissingError` **above** `except Exception` and returns without touching the DB. Marking it `failed` wasn't only a write, it was a lie — the LLM was fine, the rules are gone. Measured: halted KB + pending job → `create_app()` → db file **byte-for-byte identical**.

**(5) `status`/`coverage` were silent — this was #155 itself, on the one surface a non-web user has.** Fixed: every `halt_safe=True` command reads `resolve_policy` and prints the banner on **stderr**, exit code untouched (diagnosis stays usable and pipeable). Judgement is `resolve_policy` only — no file-existence checks. A healthy KB prints nothing; a banner that's always on is wallpaper.

**(2)** `PolicyMissingError` → `error: ...` + **rc=2**, handled in `main()` rather than in `cmd_sync`/`cmd_repair`. Putting it in `main()` is what the file's own comment already argues for ("a guard sprinkled per-command is a guard the next command will forget"), it covers `query` and anything added later — and a *redundant* per-command handler can't be mutation-tested, so I left it out.

**(3)** Guilty. Your mutation passed. Now pinned: on a halted KB the fake client is called **0 times** and job/chunk/`runs` rows are unchanged.

**(4)** You were right that (a)/(b) was a false choice — (a) alone leaves the zombie. I took both halves. The contract, stated honestly:

> A halted KB gains **no candidate fact and no evidence row — ever**. Job bookkeeping is only ever **rewound, never advanced**.

`assert_writable` moved **before** `mark_chunk_running` (so no post-deletion chunk is claimed at all), the in-flight chunk is rolled back to `pending` (job → `pending`, resumable after `policy reset --force`, **no zombie `running`**), and the `runs` row gets an honest summary instead of a blank one. The one residue I did *not* rewind is the in-flight chunk's `attempts` counter — it truthfully records that the chunk was attempted. Both final states are pinned by tests.

**(6)** Noted, not touched. Merge note is in the PR body: in `main()` the guard block must survive **after** `_config_for(args)`, and every subparser must keep its `halt_safe=`. `init`/`seed` losing it fails closed; `status`/`coverage`/`ui`/`policy reset` losing it bricks the recovery path.

---

**711 passed, 7 skipped** (was 701), `ruff check` clean, 10 new tests. Mutation: each of the 7 new guards deleted in turn → matching test **red** (table in the PR body). By hand: (a) `create_app()` on halted KB + pending job → `resumed_job_ids == []`, db bytes identical · (b) policy deleted mid-`sync` → rc=2, no traceback, job `pending`, 0 facts · (c) `status`/`coverage` → rc=0 + halt banner on stderr · (d) `policy reset --force` → `ingest` rc=0.

Also corrected the PR body's "does not accept writes" heading — you were right that it claimed more than the code delivers.

PR #182, opened 2026-07-13
fix(#176): derive the engine-input status filter from ENGINE_STATUSES

### LINKED ISSUE #176: Derive coverage's engine-status filter from ENGINE_STATUSES
## Symptom

Two definitions of "what counts as engine input" exist, and they agree only by coincidence.

- `ENGINE_STATUSES = frozenset({"confirmed", "accepted"})` (`verinote/store/db.py:21`) — used by
  `engine_fact_terms` and `toggle_review`
- a hardcoded `f.status IN ('confirmed','accepted')` in SQL (`db.py:113`, `source_fact_counts`) —
  which is what `coverage()` reports as engine coverage

Adding a status to the frozenset makes the coverage report assert a number the engine disagrees
with — the report would claim coverage the engine never granted.

## Fix direction

Derive the SQL filter from `ENGINE_STATUSES` so there is one definition. Cheap to fix now; this is
exactly the shape that becomes a wrong-assertion bug the moment someone adds a status.

_Recurrence of factlog's second-most-frequent bug class: one question, two implementations._

SeoyunL:
Closes #176

## Problem

Two pieces of code answered "what is engine input": the `ENGINE_STATUSES` constant (`db.py:22`) and `IN ('confirmed','accepted')` hard-coded in SQL. They agreed only by coincidence.

The stale line number in the issue body hid the blast radius — the literal appeared in **two** query sites, and a **third** answer lived in the web dashboard template:

| Location | Function | Consumer |
|---|---|---|
| `db.py:207` | `sources_with_counts` -> `engine_count` | `web/app.py` -> `sources.html:75` ("N confirmed") |
| `db.py:302` | `source_fact_counts` -> `engine` | `engine/coverage.py:52` -> `coverage()` -> `cli.py:296`, `web/app.py` |
| `dashboard.html:11` | template summed `confirmed + accepted` | the "engine input" card at the top of the UI |

Add a status and the coverage report asserts engine-input counts the engine does not agree with — and a user deletes facts on the strength of that report.

## Change

Factor out `_status_filter(statuses) -> (placeholders, params)`, the idiom the repo already had inline in `review_queue_ids` / `review_queue_page`:

- **Parameter binding, not string building.** Status values never enter SQL text, so the SQL-injection boundary does not rest on a "these are constants, so it's safe" invariant that will eventually break.
- **`sorted()`** fixes the parameter order a `frozenset` leaves unspecified.
- **Read at call time.** No default-arg capture, no import-time frozen SQL — otherwise the mutation tests below are impossible and the drift net disappears.
- **Empty tier rejected loudly.** SQLite answers `status IN ()` with zero rows instead of raising. An empty constant would therefore have made `coverage()` call *every* source a gap, and made `accept_review_facts_for_source` promote nothing while returning success — the human review gate silently disabled. Unreachable today, but this PR's whole claim is that the constant is the single handle; leaving its fatal input undefended contradicts that. A crash is cheaper than a quiet wrong answer.

Call sites: `sources_with_counts`, `source_fact_counts`, and `facts()` — the engine's *real* input path, which `ask.py`, `trust.py`, `query_schema.py` and `engine_fact_terms` all read through. Plus pure refactors of `review_queue_ids`, `review_queue_page`, and the one-line review-tier filter in `accept_review_facts_for_source`.

The dashboard card is now derived in the route (`sum(counts.get(s, 0) for s in ENGINE_STATUSES)`) and merely rendered by the template. `web/app.py` imports the store as a module so the tier resolves at call time rather than pinning a stale copy at import.

Deliberately untouched: the `candidate` / `needs_review` per-status count columns (different question — individual counts, not a set filter), `SET status = 'confirmed'` (a single write target, not set membership), `schema.sql` CHECK constraints (DDL, not runtime-derivable), `report_trace.py`'s separate `REVIEW_STATUSES` redefinition (same family, different constant — its own issue), and the membership of `ENGINE_STATUSES` itself. This unifies the definition; it does not change policy.

## Tests

Pinned by behaviour, not by snapshot numbers and not by string-matching the source (the string-matching-drift failure mode from the #155 review).

- **T1** (`test_coverage_engine_tier_follows_engine_statuses`): a source with only `superseded` facts is a gap with `engine_facts == 0`. `monkeypatch` `ENGINE_STATUSES |= {"superseded"}` and the same source stops being a gap with `engine_facts == 2`.
- **T2** (`test_coverage_engine_facts_match_the_engines_own_input`): `coverage.engine_facts == len(store.facts(statuses=db.ENGINE_STATUSES) for this source)` — cross-checked against the engine's own input path, asserted under both default and mutated tiers. What is fixed is the invariant, not a number.
- **T3** (`test_sources_with_counts_engine_count_follows_engine_statuses`): same mutation assertion for `engine_count`, so the Sources page cannot report a different number than coverage. Also asserts the `candidate_count` / `fact_count` columns do **not** move.
- **Dashboard** (`test_dashboard_engine_input_card_follows_engine_statuses`): parses the rendered card and asserts it tracks the tier, and equals what the engine would actually read.
- **Empty tier**: positive/negative cases across the engine and review paths (`facts`, `source_fact_counts`, `sources_with_counts`, `review_queue_page`, `accept_review_facts_for_source`), including that no fact is promoted behind our back. `statuses=None` still means "no filter", not "empty filter".
- **Consumer agreement** (`test_sources_with_counts_engine_count_matches_coverage`) now mutates too. It previously *survived* the literal revert: two consumers reading the same wrong literal still agree with each other, so agreement alone was not evidence.

**Mutation proof (run, not assumed).** Reverted both queries to the literals and the template to its inline sum:

```
FAILED tests/test_store.py::test_sources_with_counts_engine_count_follows_engine_statuses
FAILED tests/test_store.py::test_sources_with_counts_engine_count_matches_coverage
FAILED tests/test_store.py::test_engine_reads_are_refused_when_the_engine_tier_is_empty
FAILED tests/test_coverage.py::test_coverage_engine_tier_follows_engine_statuses
FAILED tests/test_coverage.py::test_coverage_engine_facts_match_the_engines_own_input
FAILED tests/test_web.py::test_dashboard_engine_input_card_follows_engine_statuses
6 failed, 127 passed
```

(The coverage failure is `assert 0 == 2` — the report still says "no engine facts" while the engine would read two.) Restored; the net catches the regression it exists to catch, on every consumer.

## Verification

- Full suite: **677 passed, 7 skipped** — no regressions. Under the default `ENGINE_STATUSES` every number is byte-identical; the production behaviour change is a refactor plus a new guard on an unreachable input.
- `verinote coverage` output unchanged on a scratch KB (`1/1 engine facts`, `0/1 ... GAP`, `coverage: 1 covered, 1 gap(s), 0 orphan(s)`).

---

## Review round 2 (justinjoy, CHANGES_REQUESTED)

The first round moved Store/coverage/dashboard to call-time lookup but left the escape hatch open: `from verinote.store import ENGINE_STATUSES` binds the frozenset **object** at import time. So the reviewer's repro was real — widen `db.ENGINE_STATUSES` and `Store.engine_fact_terms()` reads the new fact while `build_query_schema_snapshot().fact_count` stays `0` and `ask.grounding_facts()` stays empty. The deterministic engine would consume facts the planner's schema hint and the Ask/trust layers cannot see.

**Fix: first-class accessors, and no way to bind the tier at import time.**

New `verinote/store/tiers.py` — `engine_statuses()`, `review_statuses()`, `is_engine_input(status)`, `is_review_eligible(status)` — all resolving `db.ENGINE_STATUSES` / `db.REVIEW_STATUSES` at call time, all refusing an empty tier for the same reason `_status_filter` does. `verinote/store` **no longer re-exports the raw frozensets** (breaking for any importer of `verinote.store.ENGINE_STATUSES`; `db.py` keeps them as the owner). Chose the accessor route over "sprinkle `store_db.X` everywhere" because the repeated lesson here is that two places answering one question drift — an accessor cannot be cached by accident, and removing the re-export makes the old mistake unavailable rather than merely discouraged.

Consumers converted (blocking item 1, plus 2 and 3): `pipeline/query_schema.py`, `pipeline/ask.py`, `pipeline/trust.py`, `pipeline/corroboration.py` (this is what the dashboard's corroboration/conflict sections go through — item 2), `pipeline/workbench.py`, `pipeline/acceptance.py` (item 3).

Item 4: `sources.html` now labels `engine_count` **"N engine input"**, matching the dashboard card and what the number actually means (`ENGINE_STATUSES` membership, `accepted` included).

`web/app.py` and `db.py` are untouched this round — app.py already read `store_db.ENGINE_STATUSES` at call time, and keeping both files out of the diff avoids churn against #179.

### Mutation evidence (`tests/test_status_tiers.py`, 7 tests)

Every test widens `db.ENGINE_STATUSES` with `superseded` (or empties a tier) and asserts the consumer moves. Reverting the six pipeline modules to their import-time bindings (and re-adding the re-export so they still import) turns **all 7 red**, including the reviewer's exact split:

```
>       assert snapshot.fact_count == 1
E       AssertionError: assert 0 == 1
        # while store.engine_fact_terms() already returned [1]
```

- engine-input readers agree after widening: `engine_fact_terms` / `build_query_schema_snapshot` / `grounding_facts`
- trust: `engine_input` flag and distinct-source support
- corroboration + single-valued conflicts + workbench (the dashboard sections)
- acceptance: support sources / `insufficient_distinct_source_support`
- empty engine tier -> every consumer raises `ValueError` (not "zero facts")
- empty review tier -> `accept_recommendations()` / `accept_recommendation()` raise, and the fact is **not** promoted behind our back (item 3 closed)
- an AST guard: no module outside `db.py`/`tiers.py` may name the tiers in an import — the regression is an import statement, so it is guarded as one

Also new: `test_sources_page_labels_the_engine_count_by_what_it_means` (item 4 — asserts "N engine input", no "confirmed", and that the badge tracks a widened tier).

### Validation

- `.venv/bin/python -m pytest -q` -> **685 passed, 7 skipped** (677 before this round)
- `.venv/bin/ruff check .` -> All checks passed
- Merge simulation against `issue-155-loud-missing-policy` (#179): `db.py` and `web/app.py` auto-merge cleanly; one 5-line conflict in `verinote/store/__init__.py` (#179 adds `POLICY_MARKER_KEY` to the same import/`__all__` lists this PR trims) — resolve by keeping `POLICY_MARKER_KEY` and dropping the two tier names.

### DIFF
--- a/tests/test_coverage.py
+++ b/tests/test_coverage.py
@@ -1,7 +1,22 @@
 # SPDX-License-Identifier: MPL-2.0
 from verinote.engine import coverage
-from verinote.store import Store
+from verinote.store import Store, db
 from verinote.store.fact_input import structural_term
+
+
+def _engine_facts_for_source(store: Store, source_id: int) -> int:
+    """Count the facts the engine itself would read for one source.
+
+    Deliberately routed through the same status constant the engine reads
+    (`db.ENGINE_STATUSES`, looked up at call time) rather than a literal, so the
+    assertions below fix an invariant instead of a snapshot number.
+    """
+    rows = store.facts(statuses=db.ENGINE_STATUSES)
+    return len([r for r in rows if r["source_id"] == source_id])
+
+
+def _source_coverage(store: Store, tmp_path, path: str):
+    return next(s for s in coverage(store, root=tmp_path).sources if s.path == path)
 
 
 def _store(tmp_path) -> Store:
@@ -67,3 +82,62 @@ def test_orphan_when_backing_file_missing(tmp_path):
     sc = cov.sources[0]
     assert sc.is_orphan is True
     assert sc in cov.orphans
+
+
+def test_coverage_engine_tier_follows_engine_statuses(tmp_path, monkeypatch):
+    """Coverage must derive its engine tier from ENGINE_STATUSES, not from SQL.
+
+    A source whose only facts are `superseded` is a gap today. Widen
+    ENGINE_STATUSES to include `superseded` and the very same source must stop
+    being a gap — because the engine would now read those facts. Hard-coding
+    `IN ('confirmed','accepted')` back into the query makes this test fail.
+    """
+    s = _store(tmp_path)
+    path = _with_file(tmp_path, "a.txt")
+    sid = s.add_source(path)
+    s.add_fact("A", "is_a", "B", status="superseded", source_id=sid)
+    s.add_fact("C", "is_a", "D", status="superseded", source_id=sid)
+    superseded_count = 2
+
+    before = _source_coverage(s, tmp_path, path)
+    assert before.engine_facts == 0
+    assert before.is_gap
+    assert before.total_facts == superseded_count
+
+    monkeypatch.setattr(
+        db, "ENGINE_STATUSES", db.ENGINE_STATUSES | {"superseded"}
+    )
+
+    after = _source_coverage(s, tmp_path, path)
+    assert after.engine_facts == superseded_count
+    assert not after.is_gap
+    assert after.total_facts == superseded_count
+
+
+def test_coverage_engine_facts_match_the_engines_own_input(tmp_path, monkeypatch):
+    """The number coverage reports == the number of facts the engine reads.
+
+    Asserted against the engine's own input path (`facts(statuses=...)`) rather
+    than a literal, and re-checked under a mutated ENGINE_STATUSES so the two
+    definitions cannot drift apart silently.
+    """
+    s = _store(tmp_path)
+    path = _with_file(tmp_path, "a.txt")
+    sid = s.add_source(path)
+    s.add_fact("A", "is_a", "B", status="confirmed", source_id=sid)
+    s.add_fact("C", "is_a", "D", status="accepted", source_id=sid)
+    s.add_fact("E", "is_a", "F", status="needs_review", source_id=sid)
+    s.add_fact("G", "is_a", "H", status="superseded", source_id=sid)
+
+    default_cov = _source_coverage(s, tmp_path, path)
+    assert default_cov.engine_facts == _engine_facts_for_source(s, sid)
+
+    monkeypatch.setattr(
+        db, "ENGINE_STATUSES", db.ENGINE_STATUSES | {"superseded"}
+    )
+
+    mutated_cov = _source_coverage(s, tmp_path, path)
+    assert mutated_cov.engine_facts == _engine_facts_for_source(s, sid)
+    # The invariant held both times *and* the tier actually moved, so the
+    # equality above is not vacuous.
+    assert mutated_cov.engine_facts == default_cov.engine_facts + 1
--- a/tests/test_status_tiers.py
+++ b/tests/test_status_tiers.py
@@ -1,0 +1,226 @@
+# SPDX-License-Identifier: MPL-2.0
+"""One runtime definition of the status tiers, proven by mutation.
+
+`verinote.store.db` owns `ENGINE_STATUSES` / `REVIEW_STATUSES`. Every consumer
+must answer "is this fact engine input?" from that one constant, *at call time*.
+The failure this file exists to prevent is not hypothetical: consumers used to
+do `from verinote.store import ENGINE_STATUSES`, which binds the frozenset
+object at import time, so widening the tier moved `Store.engine_fact_terms()`
+(a call-time lookup) while `build_query_schema_snapshot()` and
+`ask.grounding_facts()` stayed on the old tier — the deterministic engine would
+consume facts the planner's schema hint and the Ask/trust layers cannot see.
+
+Each test below widens `db.ENGINE_STATUSES` (or empties a tier) and asserts the
+consumer moves with it. Reverting a consumer to an import-time binding turns
+these red.
+"""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+import pytest
+
+from verinote.pipeline.acceptance import accept_recommendation, accept_recommendations
+from verinote.pipeline.ask import grounding_facts
+from verinote.pipeline.corroboration import (
+    store_corroboration,
+    store_single_valued_conflicts,
+)
+from verinote.pipeline.query_schema import build_query_schema_snapshot
+from verinote.pipeline.trust import fact_trust_summary
+from verinote.pipeline.workbench import trust_workbench
+from verinote.store import Store, db
+
+
+def _store(tmp_path) -> Store:
+    store = Store(tmp_path / "kb.sqlite")
+    store.init_schema()
+    return store
+
+
+def _widen(monkeypatch) -> None:
+    """Add `superseded` to the engine-input tier, the reviewer's mutation."""
+    monkeypatch.setattr(db, "ENGINE_STATUSES", db.ENGINE_STATUSES | {"superseded"})
+
+
+def test_engine_input_readers_agree_after_the_tier_is_widened(tmp_path, monkeypatch):
+    """The reviewer's exact repro: Store saw the fact, schema and Ask did not.
+
+    A `superseded` fact is invisible to every engine-input reader today. Widen
+    the tier and all of them must see it together — otherwise the engine reasons
+    over facts the planner's schema hint and Ask never learned about.
+    """
+    store = _store(tmp_path)
+    source_id = store.add_source("sources/a.md")
+    store.add_fact("Acme", "uses", "FastAPI", status="superseded", source_id=source_id)
+
+    # Before: nobody reads it.
+    assert store.engine_fact_terms() == []
+    assert build_query_schema_snapshot(store).fact_count == 0
+    assert grounding_facts(store, question="Acme uses what?") == []
+
+    _widen(monkeypatch)
+
+    # After: everybody reads it, from the same constant, at call time.
+    assert [row["id"] for row in store.engine_fact_terms()] == [1]
+    snapshot = build_query_schema_snapshot(store)
+    assert snapshot.fact_count == 1
+    assert [r.relation.display for r in snapshot.relations] == ["uses"]
+    grounded = grounding_facts(store, question="Acme uses what?")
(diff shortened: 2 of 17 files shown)


justinjoy (changes requested):
Requesting changes for one blocking consistency issue and a few related gaps.

1. **`ENGINE_STATUSES` is still not a single runtime definition for engine-input consumers.**

This PR moves Store/coverage/dashboard pieces toward call-time `store_db.ENGINE_STATUSES` lookup, but several engine-input consumers still bind the re-exported frozenset at module import time via `from verinote.store import ENGINE_STATUSES`.

Examples:

- `verinote/pipeline/query_schema.py` imports `ENGINE_STATUSES` at module import and uses it in `_engine_facts()`.
- `verinote/pipeline/ask.py` imports `ENGINE_STATUSES` at module import and uses it in `grounding_facts()`.
- `verinote/pipeline/trust.py` imports `ENGINE_STATUSES` at module import and uses it for engine-input reads/checks.

I reproduced the split locally: after widening `verinote.store.db.ENGINE_STATUSES` to include `superseded`, `Store.engine_fact_terms()` started reading the superseded fact, but `build_query_schema_snapshot(...).fact_count` stayed `0` and `ask.grounding_facts(...)` stayed empty. That means the deterministic engine can consume facts that the planner/schema hint and Ask/trust layers do not see. This violates the PR's stated call-time lookup invariant.

Suggested fix: either import `verinote.store.db as store_db` in all status-tier consumers and read `store_db.ENGINE_STATUSES`/`store_db.REVIEW_STATUSES` at call time, or expose first-class Store APIs/accessors for engine-input and review-tier reads so callers do not pass cached tier objects around.

2. **Dashboard sections can still disagree with the headline engine-input card.**

The dashboard now derives the headline `engine_input` card from `store_db.ENGINE_STATUSES`, but the same page renders corroboration/conflict sections through `store_corroboration()` / `store_single_valued_conflicts()`, whose module still captured `ENGINE_STATUSES` at import time. Under this PR's own mutation model, the card and coverage move when `db.ENGINE_STATUSES` is widened, while corroboration/workbench can remain on the old tier.

3. **The empty-tier guard is not proven across review/acceptance consumers.**

The new `_status_filter()` guard is good for Store methods, but acceptance still imports `REVIEW_STATUSES` and `ENGINE_STATUSES` at module import time. Replacing `db.REVIEW_STATUSES` with an empty set does not necessarily make `accept_recommendations()` observe the empty tier, so the “empty tier must crash” invariant is not closed over the review/acceptance path.

4. **Minor wording issue: Sources still labels `engine_count` as confirmed.**

`verinote/web/templates/sources.html` renders the derived `engine_count` as “N confirmed”. The value already includes `accepted`, and after this PR it explicitly means “facts in `ENGINE_STATUSES`”. That label should be updated to avoid misleading users and future maintainers.

Validation I ran:

- GitHub CI is green: `ruff`, `pytest (py3.11)`, `pytest (py3.12)`, `pytest (py3.13)` all passed.
- Local targeted tests: `6 passed`.
- Local full suite: `.venv/bin/python -m pytest -q` -> `677 passed, 7 skipped`.
- Additional local probe reproduced the stale-import disagreement described above.

SeoyunL:
Thanks — the repro was exact, and it was the whole point of the PR failing on its own terms. Pushed 21cd3e1.

**1 (blocking) — fixed by removing the escape hatch, not just the usages.** New `verinote/store/tiers.py` exposes `engine_statuses()` / `review_statuses()` / `is_engine_input()` / `is_review_eligible()`, each resolving `db.ENGINE_STATUSES` / `db.REVIEW_STATUSES` at call time. `verinote.store` **no longer re-exports the raw frozensets**, so `from verinote.store import ENGINE_STATUSES` is now an ImportError rather than a footgun. I took the accessor route over `store_db.X` at each call site because the lesson this repo keeps re-learning is that two places answering one question drift; an accessor cannot be cached by accident. (Breaking for external importers of `verinote.store.ENGINE_STATUSES`; `db.py` keeps them as the owner.)

Converted: `query_schema.py`, `ask.py`, `trust.py`, `corroboration.py`, `workbench.py`, `acceptance.py`.

**2 — corroboration/conflicts now move with the card.** They go through `corroboration.py`, which was the module capturing the tier at import. `store_corroboration()` / `store_single_valued_conflicts()` / `trust_workbench()` are all pinned under the widening mutation now.

**3 — the empty-tier invariant is closed over review/acceptance.** The accessors refuse an empty tier for the same reason `_status_filter` does, so with `db.REVIEW_STATUSES = frozenset()`, `accept_recommendations()` and `accept_recommendation()` raise `ValueError` and the fact is provably not promoted behind our back.

**4 — label fixed.** `sources.html` now renders "N engine input", matching the dashboard card and what the number means (`ENGINE_STATUSES` membership, `accepted` included).

**Mutation evidence.** `tests/test_status_tiers.py` (7 tests): widen `db.ENGINE_STATUSES` with `superseded` and assert every consumer moves together. Reverting the six modules to import-time bindings (with the re-export restored so they still import) turns **all 7 red** — including your split:

```
>       assert snapshot.fact_count == 1
E       AssertionError: assert 0 == 1
        # while store.engine_fact_terms() already returned [1]
```

There is also an AST guard test: no module outside `db.py`/`tiers.py` may name the tiers in an import. The regression is an import statement, so it is guarded as one.

**Validation:** `pytest -q` -> **685 passed, 7 skipped** (677 before); `ruff check .` -> clean.

**On #179:** I kept `db.py` and `web/app.py` out of this round (app.py already read the tier at call time). Local merge simulation against `issue-155-loud-missing-policy`: both files auto-merge cleanly; the only conflict is a 5-line one in `verinote/store/__init__.py`, where #179 adds `POLICY_MARKER_KEY` to the same import/`__all__` lists this PR trims — resolve by keeping `POLICY_MARKER_KEY` and dropping the two tier names.

justinjoy (approved):
Re-reviewed the prior change requests. The stale import-time tier bindings are removed, engine/review tier access is routed through call-time accessors, dashboard/source labels are aligned, and the new mutation tests cover the reported drift. Local validation: targeted status/store/coverage/web tests, full pytest, and ruff all pass.

Issue #185, opened 2026-07-13
Reconsider active_root()'s CWD-relative ./data fallback and the unconditional global app.json write


SeoyunL:
Surfaced while closing #163 (PR #184). That PR stopped the *test suite* from repointing the user's active KB, but it did so entirely in the test harness — the production design that makes the accident easy is unchanged. Filing separately so it does not get buried under a green test PR.

## Problem

Two properties of KB selection combine badly.

**1. The fallback is CWD-relative.** `verinote/config.py`:

```python
def _default_root() -> Path:
    return Path("./data").expanduser().resolve()   # relative to the process CWD
```

`active_root()` resolves `VERINOTE_ROOT` → the global `app.json`'s `active_root` (only if `<root>/kb.sqlite` exists) → `_default_root()` → `None`. So **where you launch `verinote` from decides which knowledge base you open.** Run it from the repo and you get the repo's `data/`; run it from elsewhere and you get a different KB, or none.

**2. The global write is unconditional.** `save_active_root()` overwrites a single machine-wide `app.json` with no confirmation. Its only production caller is `_open_root` (`verinote/web/app.py`), reachable from `POST /settings/root` and `POST /kb/select` — so **one web request silently repoints the active KB for every other process on the machine**, including a CLI session the user has open in another terminal.

## Why this matters here

The active KB is where human accept/reject decisions land. Both properties mean those decisions can go to a KB the user did not intend, with no prompt and no trace. #163 was one instance of this (pytest, running in the repo root, hitting both paths); the class is larger than the test suite.

## Fix direction

Make KB selection explicit rather than ambient:

- Replace the CWD-relative fallback with something location-independent — a required `--root`/`VERINOTE_ROOT`, or an upward search for a KB marker file (the git/`pyproject` pattern), so the same command means the same KB regardless of CWD.
- Narrow the global-state write to an explicit user action. A settings POST changing machine-wide state for other running processes is surprising; at minimum it should be scoped or confirmed.

Not urgent in the sense of "corrupts data today" — but it is the mechanism behind #163, and it will keep producing that shape of bug.

SeoyunL:
Inheriting a residual from #159 (PR #190).

#159 was "stop gitignoring hand-written policy and the review database". Its headline symptom was that a routine `git clean -fdx` deletes the entire KB — every accept/reject decision and the audit log — with no git history to recover from.

[...]

PR #184, opened 2026-07-13
test(#163): sandbox HOME and CWD so no test can reach the real KB

SeoyunL:
Closes #163

## What this actually is

The test the issue points at (`test_settings_switches_active_kb_root`) **already isolates `HOME`** — it was fixed in 2bb0bae6 and is isolated on `main` today. So there is no line to fix there. What remained was the issue's fix direction: **structural containment**, so isolation is not something each test author has to remember. That is what this PR adds. The existing manual isolation is left in place as defence in depth.

## The leak paths

Two, not one:

1. **`HOME` (and friends).** `POST /settings/root` calls `save_active_root()`, which writes the platform app config from `app_config_dir()` (`verinote/config.py:63-77`). That resolves through `Path.home()` on macOS, `APPDATA`/`LOCALAPPDATA` on Windows, and `XDG_CONFIG_HOME`/`Path.home()` elsewhere — so a sandbox has to pin **all five** of `HOME`, `USERPROFILE`, `XDG_CONFIG_HOME`, `APPDATA`, `LOCALAPPDATA`. An unisolated test rewrites the developer's real `app.json` and repoints their active KB at a temp dir pytest then deletes.
2. **`./data` (found while writing this).** `active_root()` falls back to `_default_root()` = `Path("./data").resolve()` — **relative to the CWD**. pytest runs from the repo root, so any unisolated test that reaches `Config.load()` opens the repo's own `data/kb.sqlite`. Isolating `HOME` alone does not close this; the sandbox also `chdir`s off the repo and drops an ambient `VERINOTE_ROOT`.

## Change

- `tests/conftest.py`: function-scoped **autouse** fixture pinning the five home vars at a per-test throwaway home (from `tmp_path_factory`, *not* `tmp_path` — many tests use `tmp_path` as a KB root and walk it), `delenv("VERINOTE_ROOT")`, and `chdir` to a throwaway CWD. Autouse runs first, so tests that set these vars themselves still win. Plus a **session-scoped canary** that snapshots the real `app.json` and fails the run if anything wrote to it.
- `tests/env_sandbox.py`: the sandbox's shared state. Separate module because pytest imports a conftest as top-level `conftest`, so `from tests.conftest import ...` would create a *second* module object with its own copy of the state — and would break under CI's bare `pytest` invocation, where the repo root is not on `sys.path`.
- `tests/test_env_isolation.py`: three tests that verify the guard itself. They deliberately omit manual isolation and never request the fixture by name, so they depend on it *implicitly*.
[...]

PR #186, opened 2026-07-13
fix(#165): keep init/seed on a local KB and gate demo facts

### LINKED ISSUE #165: Make init and seed operate on the named KB, not the global active KB
## Symptom

`verinote init --seed`, run in a brand-new empty directory, creates **nothing there** and instead
appends 4 fictional demo facts to the user's **existing global active KB** — 2 of them with
`status="confirmed"`, so they bypass the human review gate entirely and become DuckDB engine input
immediately.

## Repro

```
$ cd /…/newproj && verinote init --seed
initialised KB at /…/precious-kb        <-- NOT newproj
  seeded demo facts
$ ls -a /…/newproj      ->  (empty)
$ verinote status       ->  facts: 5   confirmed 3   (was 1)
```

`verinote seed` has the same reach.

## Cause

`verinote/cli.py:45-72`. `cmd_init` uses `Config.load()` → `_root()` → `active_root()`
(`config.py:56-58, 99-114`), which prefers the **saved global** KB over the current directory.
The demo facts are `_DEMO_FACTS` (`cli.py:14-23`), two of which are `confirmed`.

Related false prescription: `init --help` still says "under VERINOTE_ROOT (./data)"
(`cli.py:260`) — it does not use `./data` when a saved active KB exists.

## Fix direction

`init` should create the KB it names (cwd/`./data` or an explicit path) and never silently write
into a different, already-populated KB. Seeding unreviewed `confirmed` facts into a real KB is the
more serious half — demo facts should enter as `candidate` at most.

_Recurrence of a known factlog failure: a locally-scoped command mutating global state (`init` silently switching the active KB)._

SeoyunL:
Closes #165

## The bug

`main()` built the config with `Config.load()` for every command except `ui`/`serve`, so `init` and `seed` resolved the KB through `active_root()` — i.e. the saved global `app.json` that the web KB picker writes. Running `verinote init --seed` in an empty directory scaffolded and seeded **the saved active KB**, not a local one.

Worse, two of the four `_DEMO_FACTS` were seeded as `confirmed`. `confirmed` is in `ENGINE_STATUSES`, which `pipeline/ask.py` uses to select engine input — so demo facts became engine input with no human ever reviewing them. That is the determinism-boundary violation, and it's the more serious half of this issue.

## The fix

- **`config.local_root(explicit)`** (new): explicit arg → `VERINOTE_ROOT` → `./data` relative to the CWD. It never reads `read_app_config()`/`active_root()`. `active_root()`/`_root()`/`_default_root()` are unchanged.
- **`init` / `seed` take an optional `[root]` positional.** `main()` routes only those two through `Config.for_root(local_root(args.root))`; every other command keeps its existing `Config.load()` path, so there is no regression in global KB selection.
- **`seed` no longer creates a KB implicitly.** With no `kb.sqlite` at the target it exits 1 with `no KB at <root>; run \`verinote init\` first`.
- **Demo facts are back inside the human gate**: the two `confirmed` entries are now `candidate`. Rather than trusting the literal, a test enforces the invariant — `{status for _DEMO_FACTS} & ENGINE_STATUSES == set()` — and a second test asserts `store.facts(statuses=ENGINE_STATUSES) == []` after `init --seed`.
- Help text corrected (it claimed `under VERINOTE_ROOT (./data)`); README section on scaffolding updated.

`init` still does **not** call `save_active_root()` — it reads no global state and writes none.

## Out of scope (deliberately)

The `./data` CWD-relative fallback inside `active_root()`, the unconfirmed global `app.json` write in `web/app.py`, and the `Config.load()`/`load_for_ui()` global-selection design are untouched — that is #185.

## ⚠️ Breaking behaviour change (intended)

Anyone who had a saved active KB and ran `verinote init` from elsewhere now gets a **new KB in `./data` under the current directory** instead of silently reusing the saved one. The old behaviour was the bug. To target an existing KB, name it explicitly: `verinote seed <path>`, `verinote init <path>`, or `VERINOTE_ROOT=<path> verinote init`.

## Verification

- `.venv/bin/python -m pytest tests -q` → **677 passed, 7 skipped** (669 before; 8 new tests, no existing test changed).
- `ruff check` → All checks passed.
- Smoke run in an empty dir with a saved active KB present: `init --seed` created `./data/kb.sqlite` locally, the saved KB stayed byte-identical, and the seeded KB holds 3 `candidate` + 1 `needs_review` and **zero** engine-status facts.

New tests: saved-active-KB repro (byte-for-byte immutability of the existing KB), explicit-root `init`, `VERINOTE_ROOT` still wins, demo-status ∩ `ENGINE_STATUSES` == ∅, no engine facts after `init --seed`, `seed` with no KB → rc=1 creating nothing, `seed <root>` targeting, and help text no longer lying.

Each new test controls its own env (`VERINOTE_ROOT` deleted; `HOME`/`XDG_CONFIG_HOME`/`APPDATA` pointed at `tmp_path`) and builds the saved-KB fixture by calling `config.save_active_root()` rather than duplicating the platform branching. `tests/conftest.py` is untouched, so this stays compatible with the autouse sandbox in #184.

---

## Review round 2 (all six items closed)

**1. `init` no longer misdirects you to `verinote status` [WARNING].** The seeded-KB line said ``run `verinote status` ``, but `status`/`sync`/`ask`/`verify` still resolve through `Config.load()` → `active_root()`, so with a saved active KB that command showed a *different* KB. The false hint is gone; `init` now always prints the root it actually created as usable input:

```
initialised KB at /tmp/x/mine
  db: /tmp/x/mine/kb.sqlite
  policy: /tmp/x/mine/policy/logic-policy.dl
  seeded demo facts
  use this KB: VERINOTE_ROOT=/tmp/x/mine verinote status
```

`active_root()` itself is untouched (#185 boundary) and `cmd_init` still never calls `save_active_root()` — making `init` grab the global pointer would *be* the failure mode this issue is about. A test pins the hint to `cfg.root`.

**2. `seed` detects an unusable `kb.sqlite` [WARNING].** The guard was `db_path.is_file()`, so a 0-byte file let `_store()` create a schema (violating "only fills an existing KB") and a corrupt file raised a raw `sqlite3.DatabaseError` traceback. New `_kb_schema_problem()` opens the DB read-only and looks for the `facts` table; both cases now exit 1 with a recovery path (`... is not a verinote KB (no \`facts\` table); move it aside and run \`verinote init <root>\``) and the file is left byte-identical.

**3. `init <existing-file>` exits 1 instead of tracebacking [WARNING].** `mkdir(parents=True, exist_ok=True)` still raises `FileExistsError` when a *file* occupies the root path — a surface the new `[root]` positional opened. Guarded with a message, rc=1.

**4. A blank root is rejected, not swallowed [INFO].** `local_root()` used `if explicit:`, so `verinote init ""` fell back to `VERINOTE_ROOT`/`./data` and created a KB somewhere the caller never named. Now `if explicit is not None:`, and a blank/whitespace string raises `ValueError` → rc=1 in `main()`. `VERINOTE_ROOT="   "` behaviour is unchanged (that is `active_root()`, i.e. #185).

**5. The demo-status gate can't pass vacuously [INFO].** Added `assert demo_statuses` and `assert ENGINE_STATUSES` before the disjointness check.

**6. Regression guard for the global pointer [INFO].** `test_init_seed_in_empty_dir_ignores_saved_active_kb` now also asserts `config.read_app_config()["active_root"] == str(existing)`, so a future "helpful" `save_active_root(new)` inside `cmd_init` fails loudly instead of sliding through a bytes-only check.

### Known limitation (by design, tracked in #185)

The KB `init` creates does **not** become the active one. To use it you must pass `VERINOTE_ROOT=<root>` or select it in the UI — which is exactly what the new hint and the README now say. Fixing that properly means changing how the active KB is selected, which is #185's territory, not this PR's.

### Verification

- `.venv/bin/python -m pytest -q` → **683 passed, 7 skipped** (677 before; 6 new tests, no existing test's intent changed).
- `.venv/bin/ruff check verinote tests` → All checks passed.
- **Mutation evidence** — each new guard was removed in turn and the matching test went red, so none of them are vacuous: drop the non-dir guard → 1 failed; drop the seed schema guard → 2 failed; revert `local_root` to `if explicit:` → 2 failed; restore the false `status` hint → 1 failed; make `cmd_init` call `save_active_root()` → 1 failed; empty `_DEMO_FACTS` → 1 failed.
- End-to-end run under a fake `HOME` with a *different* KB saved as active: `init --seed` created only the named root, `VERINOTE_ROOT=<root> verinote status` read it back (4 facts), empty/corrupt `kb.sqlite` and a file-as-root and `init ""` each exited 1 changing nothing, and `app.json` still pointed at the original KB.

### Follow-up, not fixed here

Running `seed` twice duplicates the demo facts (8 rows after two runs) — pre-existing, out of scope for this PR, and worth its own issue.

---

## Follow-up fix: unescaped KB path in the SQLite URI (review round 2)

The schema guard added above shipped with a bug of its own, caught in review. `_kb_schema_problem()` built its probe URI by interpolation:

```python
conn = sqlite3.connect(f"file:{db_path}?mode=ro", uri=True)
```

A KB root containing `?` or `#` truncates that URI at the metacharacter, with two consequences (both measured on a real KB at `/tmp/X/weird?dir/`):

1. **False rejection.** The probe read a truncated path, found no `facts` table, and `seed` refused a perfectly healthy KB with rc=1.
2. **Worse: an unnamed write.** With the URI cut short, `mode=ro` was no longer parsed as a query parameter, so SQLite opened **read-write** and created a stray empty file (`/tmp/X/weird`) in the *parent* directory — writing to a location the user never named. That is a miniature of the exact behaviour this PR exists to forbid.

**Fix:** build the URI with `Path.as_uri()`, which percent-encodes both characters, so `mode=ro` survives and the probe can never create a file.

```python
uri = f"{db_path.resolve().as_uri()}?mode=ro"
conn = sqlite3.connect(uri, uri=True)
```

**Regression tests (2 new).** `test_kb_schema_probe_survives_a_uri_metacharacter_in_the_path` drives the probe directly with a `?` root; `test_seed_accepts_a_kb_whose_path_holds_a_uri_metacharacter` drives full `init` → `seed` with a `#` root. Each asserts both symptoms: (a) the healthy KB is not misreported, and (b) a before/after listing of the parent directory is unchanged, so no stray file appears.

**Mutation evidence:** reverting the fix to the interpolated URI turns **both** new tests red (`2 failed`); restoring it returns them to green (`2 passed`). Neither test is vacuous.

**Verification:** `.venv/bin/python -m pytest -q` → **685 passed, 7 skipped**; `.venv/bin/ruff check` → All checks passed.

### Second follow-up candidate (pre-existing, not fixed here)

While testing the `?` path I found that the **DuckDB** fact-term store splits paths at `?` as well, independently of this PR: seeding a KB whose root contains `?` dies with `Cannot open file ".../weird.wal?dir/facts.duckdb"`. This reproduces on the commit *before* this fix, so it is not a regression from this branch — it deserves its own issue. (It is why the end-to-end test above uses `#` rather than `?`.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,14 @@ verinote init /path/to/kb     # a named root
 verinote seed /path/to/kb     # demo facts into an existing KB
 ```
 
+Creating a KB does not make it the active one. Every other command still reads
+the saved active KB, so to work with the KB you just created either point
+`VERINOTE_ROOT` at it or select it in the UI:
+
+```bash
+VERINOTE_ROOT=/path/to/kb verinote status
+```
+
 Seeded demo facts land as `candidate`/`needs_review`, never as engine input —
 demo data has to pass through human review like anything else.
 
--- a/README.md
+++ b/README.md
@@ -69,11 +69,18 @@ tests, and one-off launches.
 VERINOTE_ROOT=/path/to/kb verinote ui
 ```
 
-You can also scaffold a KB explicitly:
+You can also scaffold a KB explicitly. `init` and `seed` are *local* commands:
+they never write to the saved active KB. They target the root you name, else
+`VERINOTE_ROOT`, else `./data` in the current directory.
 
 ```bash
-verinote init      # uses VERINOTE_ROOT, the saved active KB, or ./data
+verinote init                 # ./data here (or $VERINOTE_ROOT if set)
+verinote init /path/to/kb     # a named root
+verinote seed /path/to/kb     # demo facts into an existing KB
 ```
+
+Seeded demo facts land as `candidate`/`needs_review`, never as engine input —
+demo data has to pass through human review like anything else.
 
 DuckDB is a core dependency because it powers verification. The `analytics` extra is
 kept as a compatibility no-op; analytics uses the same DuckDB dependency. The
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,10 +1,15 @@
 # SPDX-License-Identifier: MPL-2.0
+from pathlib import Path
+
+import pytest
+
 import verinote.cli as cli
+from verinote import config
 from verinote.engine import DEFAULT_POLICY
 from verinote.llm.base import ExtractedFact, LLMError
 from verinote.pipeline.ingest import register_converter
 from verinote.pipeline.query_intent import parse_query_intent
-from verinote.store import Store
+from verinote.store import ENGINE_STATUSES, Store
 from verinote.store.fact_input import structural_term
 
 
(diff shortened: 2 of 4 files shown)


justinjoy (changes requested):
리뷰 결과 수정이 필요합니다.

1. **병합 후 테스트 collection 단계에서 오류가 발생합니다.** PR의 `tests/test_cli.py:13`이 `from verinote.store import ENGINE_STATUSES, Store`를 추가했는데, 최신 `main`에서는 `ENGINE_STATUSES`를 더 이상 `verinote.store`에서 직접 가져올 수 없습니다. 실제 병합 worktree에서 `pytest tests/test_cli.py tests/test_status_tiers.py -q`를 실행하면 `ImportError: cannot import name 'ENGINE_STATUSES' from 'verinote.store'`로 인해 테스트 수집이 중단됩니다. 테스트는 `verinote.store.db.ENGINE_STATUSES` 또는 call-time accessor 쪽으로 갱신해야 합니다.

2. **`verinote init <root>`가 기존 corrupt `kb.sqlite`에서 처리되지 않은 예외를 노출합니다.** `verinote/cli.py:100`에서 바로 `_store(cfg)`를 열고, `_store()`가 `Store.init_schema()`를 호출하면서 corrupt DB에 대해 `sqlite3.DatabaseError: file is not a database`가 그대로 올라옵니다. 이번 PR은 `seed`의 corrupt/empty DB는 사용자 메시지로 처리하지만, “unusable init roots”를 다루는 변경 범위라 `init`도 같은 방식으로 rc=1과 복구 안내를 내야 합니다.

참고: `seed` 반복 실행 시 demo facts가 중복 삽입되는 한계도 재현됐지만, PR 본문에서 follow-up 후보로 명시되어 있어 이번 차단 항목으로 보지는 않았습니다.

검증:
- PR branch 자체: `.venv/bin/python -m pytest tests/test_cli.py -q` -> `39 passed`
- PR branch 자체: `.venv/bin/ruff check verinote/cli.py verinote/config.py tests/test_cli.py README.md` -> pass
- GitHub CI: ruff, pytest py3.11/3.12/3.13 모두 pass
- 최신 `main` + PR #186 병합 worktree: `pytest tests/test_cli.py tests/test_status_tiers.py -q` -> collection error

justinjoy (approved):
이전 리뷰에서 지적한 사항이 반영되었고, 최신 main과 병합한 결과에서도 관련 테스트 및 전체 테스트가 통과했습니다.

PR #190, opened 2026-07-13
fix(#159): stop ignoring hand-written policy; anchor generated KB artifacts

### LINKED ISSUE #159: Stop gitignoring hand-written policy and the review database

SeoyunL:
Closes #159

## What was actually broken

The real bug is the **unanchored `*.dl` glob**. Gitignore extension globs match at *any* depth, so `*.dl` silently ignored every hand-written policy, test fixture and doc example in the tree.

That is not cosmetic. It is a compound failure: a policy that never gets committed makes `load_policy()` return `None`, verinote **falls back to the shipped default policy**, and the check still reports **"consistent"**. The gate looks green while the KB's own review rules are simply absent.

Same class of bug for `*.sqlite` / `*.duckdb`. I narrowed those too rather than deferring them: `config.py` always names the DB `kb.sqlite` and the sidecar `facts.duckdb`, so anchoring by path loses no real coverage, while the bare globs would swallow any test-fixture DB the same way `*.dl` swallowed policy. Fixing one member of the family and leaving two behind would just reopen the issue later.

Generated artifacts are now ignored by **KB-layout path**, never by bare extension:

```
**/facts/query.dl
**/facts.duckdb         (+ .wal, .tmp/)
**/kb.sqlite            (+ -wal, -shm)
```

Both stores' sidecars are pinned: DuckDB leaves a `.wal` on unclean shutdown and spills to `.tmp/`, exactly as SQLite leaves `-wal`/`-shm`. Anchored paths do not glob those in for free.

`/data/` (the user KB) stays ignored. `git ls-files 'data/'` is still empty.

## The issue's fix direction is wrong — disproved by experiment

The issue asks to "at minimum un-ignore the KB so `git clean -x` can't destroy it." **That does not work.** `git clean -x` removes *untracked* files **regardless of ignore status**; only *tracked* files survive. With the KB un-ignored but still untracked:

```
git clean -ndx  ->  Would remove data/          # still destroyed
git clean -nd   ->  Would remove data/policy/   # newly destroyed (ignore used to protect this)
```

So un-ignoring does not stop `-fdx` and it *strips* the ordinary `clean -fd` protection that ignoring was already giving. It is a **net loss against its own goal**, so I did not adopt it.

(Trap for anyone retrying this: a naive re-include like `!data/policy/logic-policy.dl` is a silent no-op — git does not descend into an excluded directory.)

A user KB is **user data, not a repo artifact**, so it cannot be committed either. **`clean -fdx` safety is therefore unreachable inside #159**; it is owned by #185 (move the default root out of the working tree). **This PR fixes the misclassification; it does not make `git clean -fdx` safe.**

## Red evidence (two distinct variants — do not conflate)

`tests/fixtures/policy/sample-policy.dl` is a real hand-written policy tracked in the repo. On main it is matched by `*.dl`, so it **cannot even be `git add`-ed without `-f`** — which is what makes the lock non-vacuous.

The suite goes red in two different ways, and the counts differ. Measured, not estimated:

**(a) revert `.gitignore` to main, keep the fixture -> 6 failed, 12 passed**

```
FAILED test_hand_written_sources_are_not_ignored[tests/fixtures/policy/sample-policy.dl]
FAILED test_hand_written_sources_are_not_ignored[docs/examples/logic-policy.dl]
FAILED test_hand_written_sources_are_not_ignored[verinote/policy/logic-policy.dl]
FAILED test_hand_written_sources_are_not_ignored[some/other/kb/policy/logic-policy.dl]
FAILED test_hand_written_sources_are_not_ignored[tests/fixtures/kb/facts.sqlite]
FAILED test_hand_written_sources_are_not_ignored[tests/fixtures/kb/sample.duckdb]
6 failed, 12 passed
```

**(b) revert `.gitignore` **and** delete the fixture -> 7 failed, 11 passed**

The 7th (`test_sample_policy_fixture_is_actually_tracked`) is the non-vacuity guard; it only goes red once the fixture itself is gone. An earlier revision of this PR body reported "7 failed" for the gitignore-only case, which mixed the two variants. Corrected above.

**(c) mutation: drop the two new DuckDB sidecar rules -> 2 failed, 16 passed**

```
FAILED test_generated_kb_artifacts_stay_ignored[some/other/kb/facts.duckdb.wal]
FAILED test_generated_kb_artifacts_stay_ignored[some/other/kb/facts.duckdb.tmp/spill-0.tmp]
2 failed, 16 passed
```

After, all variants restored: `18 passed`.

The lock asserts **both directions** via `git check-ignore`'s exit status (0 = ignored, 1 = not) — sources not-ignored *and* artifacts ignored. One direction alone would be vacuous: emptying `.gitignore` passes half of it, ignoring everything passes the other half.

The fixture is **not a dead file**: `test_verify_loads_hand_written_fixture_policy` loads it into a KB for real and asserts a `warn_fixture_policy_loaded` finding that the shipped default policy cannot produce — pinning that the KB's own policy was used, not the fallback.

## Docs

`.gitignore` and README now name what is **irreplaceable**: `kb.sqlite` (every accept/reject decision + audit log), `policy/logic-policy.dl`, and `policy/relation-aliases.md` / `policy/typed-relations.md` (hand-edited inputs the issue omitted — they don't match `*.dl` but die with `/data/` all the same). README adds: keep the KB outside the tree, backups are the user's responsibility, and the `git clean -fdx` hazard with a link to #185.

## Review round 2 — all 5 findings closed

**1. [CRITICAL] The README prescribed a command that does not exist.** `verinote init <path>` (positional) is introduced by #186; on this branch `init` takes no positional argument, so the single remedy this PR offered a user **could not be run**. Since this PR's deliverable is essentially documentation, its escape hatch being inert was the whole product failing.

Fixed to the form that works **today**, verified by running it on this branch alone (origin/main + this PR):

```
$ VERINOTE_ROOT=~/verinote-kb-smoke verinote init
initialised KB at /Users/seoyun/verinote-kb-smoke
  db: /Users/seoyun/verinote-kb-smoke/kb.sqlite
  policy: /Users/seoyun/verinote-kb-smoke/policy/logic-policy.dl
exit=0
```

No dependency on #186 is stated, so the two PRs merge in either order. Confirmed on the merged state (main + #186 + this PR): **both** `VERINOTE_ROOT=... verinote init` and `verinote init <path>` succeed, no conflict, **704 passed, 7 skipped**.

**2. [WARNING] `policy/logic-policy.dl` is scaffolded, not hand-written.** `cli.py` writes it from `DEFAULT_POLICY` at `init`, so the README table calling it "hand-written" was inaccurate. Corrected to **"scaffolded by `init`, then hand-edited"**. Also surfaced the real limitation the reviewer measured: a KB kept **inside the repo tree at a non-`data/` path** exposes its policy as untracked, so a blind `git add -A` commits it. Deliberately **documented, not ignored** — an ignore rule here would reintroduce the exact bug this PR exists to fix.

**3. [WARNING] DuckDB sidecars were missing.** `facts.duckdb.wal` and `facts.duckdb.tmp/` were not ignored while SQLite's `-wal`/`-shm` were listed — an inconsistency in a PR whose entire purpose is enumerating generated files correctly. Both added and locked in `ARTIFACT_PATHS`, with mutation red evidence above (variant c).

**4. [INFO] Two `.md` assertions were permanently vacuous.** No pattern in `.gitignore` — before or after — has ever threatened `.md`, so those two assertions pass even against the pre-fix `.gitignore`. Decoration, not a regression lock. **Removed**, with a comment recording why.

**5. [INFO] Red-evidence counts conflated two variants.** Corrected and split above.

**Scope held:** `verinote/**` diff is still **0 lines**; no file owned by #184 or #186 was touched.

## Verification

- `pytest -q` -> **688 passed, 7 skipped** (this PR alone)
- merged with #186 -> **704 passed, 7 skipped**, no conflict
- `ruff check` -> **All checks passed!**
- README's prescribed command **executed on this branch** -> exit 0 (above)
- `git check-ignore -q tests/fixtures/policy/sample-policy.dl` -> exit **1**, and it appears in `git ls-files`
- `git check-ignore -q data/facts/query.dl` -> exit **0**
- `git ls-files 'data/'` -> empty
- **Zero code change**: `git diff origin/main -- verinote/` is 0 lines

---

### Follow-up: rationale rewritten to survive #179 (`59008a1`)

Four places explained *why* the ignore rules had to change, and all four argued
from engine behaviour: a missing policy falls back to the shipped default and
the check **still reports "consistent"**. That is true today and false the moment
sibling PR #179 (issue #155) lands `PolicyMissingError` — false in the very
comments that justify this PR, with a green suite, because they are only prose.

Rewritten to argue from the loss that holds in both worlds: an ignored policy is
never `git add`-ed without `-f`, nothing in this repo can regenerate hand-written
rules, so they die with the working copy — whether verinote then shrugs (now) or
refuses to run (#155). #155 is *referenced*, never depended on, so the wording
stays true if it never merges.

| Location | Reader |
|---|---|
| `.gitignore` | why these globs are anchored |
| `tests/fixtures/policy/sample-policy.dl` | why this `.dl` is tracked |
| `tests/test_gitignore.py` | why the ignore rules are pinned, not the engine |
| `tests/test_verify.py` | why a *present* policy is the thing under test |

Comments and docstrings only: no ignore pattern, no assertion condition, and no
line under `verinote/**` changed. 688 passed, `ruff check` clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,8 +17,8 @@ venv/
 #   data/policy/relation-aliases.md, data/policy/typed-relations.md
 #                               hand-written policy inputs
 # Back these up yourself, or keep the KB outside the working tree entirely
-# (`verinote init <path>` / VERINOTE_ROOT). See README, "Your KB is not a repo
-# artifact". `git clean -fdx` deletes all of it.
+# (`VERINOTE_ROOT=~/verinote-kb verinote init`). See README, "Your KB is not a
+# repo artifact". `git clean -fdx` deletes all of it.
 /data/
 
 # Generated engine artifacts, anchored to the KB layout.
@@ -29,8 +29,15 @@ venv/
 # miss: an uncommitted policy makes `load_policy()` return None, verinote falls
 # back to the shipped default policy, and the check still reports "consistent".
 # Match generated *paths*, never bare extensions. See tests/test_gitignore.py.
+#
+# Engine sidecars are listed explicitly for both stores: DuckDB leaves a `.wal`
+# behind on an unclean shutdown and spills to `.tmp/`, exactly as SQLite leaves
+# `-wal`/`-shm`. Anchored paths do not glob these in for free, so omitting them
+# would leak a generated file into `git status`.
 **/facts/query.dl
 **/facts.duckdb
+**/facts.duckdb.wal
+**/facts.duckdb.tmp/
 **/kb.sqlite
 **/kb.sqlite-wal
 **/kb.sqlite-shm
--- a/.gitignore
+++ b/.gitignore
@@ -8,13 +8,32 @@ dist/
 .venv/
 venv/
 
-# verinote local state (a KB's data is local, like a working dir)
-*.sqlite
-*.sqlite-wal
-*.sqlite-shm
-*.duckdb
+# verinote local state (a KB's data is local, like a working dir).
+#
+# The default KB lives here. It is USER DATA, not a repo artifact, so it is
+# never committed — and nothing in it is reproducible from this repo:
+#   data/kb.sqlite              every accept/reject decision + the audit log
+#   data/policy/*.dl            hand-written logic policy
+#   data/policy/relation-aliases.md, data/policy/typed-relations.md
+#                               hand-written policy inputs
+# Back these up yourself, or keep the KB outside the working tree entirely
+# (`verinote init <path>` / VERINOTE_ROOT). See README, "Your KB is not a repo
+# artifact". `git clean -fdx` deletes all of it.
 /data/
-*.dl
+
+# Generated engine artifacts, anchored to the KB layout.
+#
+# These are ANCHORED ON PURPOSE. Bare extension globs (`*.dl`, `*.sqlite`,
+# `*.duckdb`) match at any depth, so they silently ignored hand-written policy,
+# test fixtures and doc examples anywhere in the tree. That is not a cosmetic
+# miss: an uncommitted policy makes `load_policy()` return None, verinote falls
+# back to the shipped default policy, and the check still reports "consistent".
+# Match generated *paths*, never bare extensions. See tests/test_gitignore.py.
+**/facts/query.dl
+**/facts.duckdb
+**/kb.sqlite
+**/kb.sqlite-wal
+**/kb.sqlite-shm
 
 # tooling
 .pytest_cache/
--- a/.gitignore
+++ b/.gitignore
@@ -11,11 +11,21 @@ venv/
 # verinote local state (a KB's data is local, like a working dir).
 #
 # The default KB lives here. It is USER DATA, not a repo artifact, so it is
-# never committed — and nothing in it is reproducible from this repo:
+# never committed — and the only file in it this repo can regenerate is
+# `facts/query.dl`:
 #   data/kb.sqlite              every accept/reject decision + the audit log
+#   data/facts.duckdb           the canonical fact terms. NOT rebuildable from
+#                               kb.sqlite: the SQLite columns are display mirrors,
+#                               so a lost sidecar re-types compounds as strings
+#                               (#156). Ignored because it is user data, not
+#                               because it is disposable.
+#   data/sources/, data/artifacts/
+#                               the ingested documents and their extracted text
 #   data/policy/*.dl            hand-written logic policy
 #   data/policy/relation-aliases.md, data/policy/typed-relations.md
 #                               hand-written policy inputs
+#   data/policy/prompts/, data/config.json
+#                               hand-written prompt overrides and settings
 # Back these up yourself, or keep the KB outside the working tree entirely
 # (`VERINOTE_ROOT=~/verinote-kb verinote init`). See README, "Your KB is not a
 # repo artifact". `git clean -fdx` deletes all of it.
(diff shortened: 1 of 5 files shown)


justinjoy (approved):
검토 결과 머지를 막을 기능 결함은 발견하지 못했습니다. 테스트 격리 개선은 후속 PR로 처리하겠습니다.

PR #195, opened 2026-07-13
fix(#174): stop fact values forging ERROR/WARN lines in the engine report

### LINKED ISSUE #174: Escape fact values rendered into the report so they cannot forge ERROR lines
## Symptom

Fact text flows unescaped into the logic-check report, so a fact value containing newlines/tabs can
render a **visually convincing fake `ERROR` line**:

```
rep.answers -> 'q1: mathematician, pioneer\nERROR forged: the_kb_is_broken\tTAB-SHIFTED'
```

Jinja autoescapes, so this is not XSS — but the report is rendered inside
`<pre class="report">{{ rep.text }}</pre>` (`report.html:29`), where the forged line is
indistinguishable from a real engine finding.

## Cause

`verinote/engine/duckdb_backend.py:436-439` — `_render_output_term` returns `StringLit.value`
**raw**, while `_render_fact_input` (line 424) correctly uses the escaping `render_term`
(`terms.py:125-133`, which escapes `\ " \n \r \t`). The escaping helper exists; the report path just
doesn't use it.

## Fix direction

Route report rendering through the same escaping used for fact input. The general rule from factlog:
define "fields the caller can influence" as a category and neutralize all of them — guarding one
field means the next one is the hole.

Note the fact→Datalog path itself is sound (parameterized inserts, JSON-encoded terms), so this is
display-layer only.

SeoyunL:
Closes #174

## Problem

A fact value could forge a line in the engine report. Escaping `\n`/`\r` alone is not enough: `str.splitlines()` also breaks on VT, FF, FS, GS, RS, NEL, U+2028 and U+2029, so a value like `broken<VT>ERROR forged: ...` still produced an extra `ERROR ` line the engine never derived.

## Fix

`escape_string_value` in `verinote/engine/terms.py` is now the single owner of neutralization, used by both the DuckDB backend and the legacy wirelog renderer, so the two paths cannot drift apart. It escapes:

- **Cc / Zl / Zp** — a category whitelist, not a blacklist. Proven by an exhaustive 0..0x10FFFF scan to be a superset of every code point `str.splitlines()` breaks on (`{0a,0b,0c,0d,1c,1d,1e,85,2028,2029}` — all of them Cc/Zl/Zp). Also covers NUL and ESC.
- **the nine bidi controls** (U+202A–202E, U+2066–2069). They break no line, but they reverse the visual order of the text around them, so a value could make a report line *read* as a claim the engine never derived. A report is a trust artifact, so these are neutralized too.

Nothing else. In particular **all other `Cf` characters are preserved**: ZWJ (U+200D), ZWNJ (U+200C) and soft hyphen carry meaning — they join emoji sequences and are required spelling in Persian and Indic scripts. An earlier revision of this branch escaped the whole `Cf` category, which corrupted `👨‍👩‍👧`, Persian `می‌خواهم` and Devanagari `क‍ष`. That violated the README promise that "Subjects and objects preserve the source document's language and named-entity spelling", and it was never needed for #174 — those characters cannot start a line. Fixed, and pinned with tests.

Option C is kept: report values are **not** quoted (an answer surface stays `London`, not `"London"`).

## Behavior change worth calling out

The term parser now **accepts `\uXXXX` / `\UXXXXXXXX` escapes** in quoted strings. This is required for the fix to be lossless: what we render, we must be able to read back (round-trip verified over all of Unicode, 0 failures). It is a change beyond the strict scope of #174, so, explicitly:

- **Backward compatible**: existing `.dl`/`.json`/`.md` data contains 0 literal `\u` sequences, so nothing that parsed before parses differently now.
- **Not a new capability**: raw control characters could already be placed in a quoted string before this PR. Display paths all route through the escape, and the one raw consumer (`web/app.py:314`) is a Jinja-autoescaped input attribute.
- Malformed escapes are rejected as `TermParseError` (not a bare `ValueError` from `chr()`, which callers' `except TermParseError` would not catch): lone surrogates, short/non-hex digit runs, and code points above U+10FFFF.

## Evidence

- **Suite**: `781 passed, 7 skipped`. `ruff check` — all checks passed. **0 existing test expectations modified** (the only non-addition line in the test diff is a docstring).
- **Threat model re-verified after narrowing** (0x110000 exhaustive scan): `splitlines()` split set = `{0a,0b,0c,0d,1c,1d,1e,85,2028,2029}`, categories `{Cc, Zl, Zp}`, characters left unescaped: **none**; none of them renders as more than one line. The 12 forging characters (LF CR VT FF FS GS RS NEL U+2028 U+2029 NUL ESC) are blocked in **both** the DuckDB backend and the wirelog renderer.
- **Precision**: byte-identical over all of Unicode = **1,111,987** code points; escaped = 77. Everything escaped beyond Cc/Zl/Zp is exactly the nine bidi controls — nothing else is touched. Round-trip over every code point: 0 failures.
- **Mutation (each reverted change goes red)**:
  - restore the `Cc/Cf/Zl/Zp` blanket → **7 failed** (ZWJ/ZWNJ/emoji/soft-hyphen preservation)
  - drop the bidi code points → **10 failed**
  - delete the surrogate/range guard in `parse_unicode_escape` → **4 failed**
  - back to the old LF/CR/TAB blacklist → **37 failed**

Correcting figures reported earlier on this branch: the whitelist-vs-blacklist mutation is **26 red**, not 8, and byte-identity under the previous (over-broad) escape set was 1,111,826 — now 1,111,987 with the joiners preserved.

### DIFF
--- a/tests/test_duckdb_backend.py
+++ b/tests/test_duckdb_backend.py
@@ -428,3 +428,113 @@ def test_duckdb_backend_rejects_policy_relation_facts():
 
     assert rep.ok is False
     assert "relation facts must come from SQLite engine input" in rep.text
+
+
+_NOTE_POLICY = (
+    ".decl relation(subject: symbol, rel: symbol, object: symbol)\n"
+    ".decl error_note(subject: symbol, note: symbol)\n"
+    'error_note(S, N) :- relation(S, "note", N).\n'
+)
+_ANSWER_QUERY = (
+    ".decl answer_q1(value: symbol)\n"
+    'answer_q1(V) :- relation("Widget", "located_in", V).\n'
+)
+
+
+def _report_body(rep) -> str:
+    """The finding body, without the debug echo of the policy/query/fact input."""
+    return rep.text.split("\n\n--- policy input ---")[0]
+
+
+def test_duckdb_backend_fact_values_cannot_forge_report_lines():
+    _duckdb()
+    rep = run_check_duckdb(
+        [
+            {
+                "subject": "Widget",
+                "relation": "note",
+                "object": "broken\nERROR forged: Gadget is unusable",
+            },
+            {"subject": "Gizmo", "relation": "note", "object": "missing"},
+        ],
+        policy_dl=_NOTE_POLICY,
+    )
+
+    body = _report_body(rep)
+    claimed = [
+        line
+        for line in body.splitlines()
+        if line.startswith("ERROR ") or line.startswith("WARN ")
+    ]
+    assert rep.errors == 2
+    assert len(claimed) == rep.errors + rep.warnings
+    assert len(rep.findings) == 2
+
+
+def test_duckdb_backend_escapes_control_characters_in_answers():
+    _duckdb()
+    rep = run_check_duckdb(
+        [
+            {
+                "subject": "Widget",
+                "relation": "located_in",
+                "object": "alpha\nbeta\tgamma\rdelta",
+            }
+        ],
+        query_dl=_ANSWER_QUERY,
+    )
+
+    answer = rep.answers[0]
+    assert "\n" not in answer
+    assert "\t" not in answer
+    assert "\r" not in answer
+    assert answer == "q1: alpha\\nbeta\\tgamma\\rdelta"
+
+
+def test_duckdb_backend_escaping_is_lossless():
+    _duckdb()
+    rep = run_check_duckdb(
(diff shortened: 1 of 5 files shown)


justinjoy (approved):
검토 중 확인한 legacy report text surface의 남은 escape 이슈는 별도 후속 이슈로 추적하겠습니다. 현재 PR은 CI가 통과했고 merge state가 clean이므로 머지합니다.

PR #197, opened 2026-07-13
fix(#188): stop interpolating the KB path into ATTACH, and fail fast on a '?' in the KB root

### LINKED ISSUE #188: A '?' or '#' in the KB path breaks the DuckDB fact-term store
Found while fixing #165 (PR #186), where the same class of bug bit the new SQLite schema probe. Reproduced on `origin/main` — this is **not** a regression from that PR.

## Symptom

With a KB root whose path contains a `?` (legal on macOS and Linux), the DuckDB sidecar fails to open — the path is truncated at the metacharacter:

```
Cannot open file ".../weird.wal?dir/facts.duckdb"
```

So `seed`, and anything else that touches fact terms, dies on a KB the filesystem is perfectly happy with.

## Cause

A path is being interpolated into a string that DuckDB parses as a URI (or otherwise splits on `?`), rather than being passed as a literal path or percent-encoded. The same mistake, in `verinote/cli.py`'s SQLite probe, additionally caused a **write to an unnamed location**: `mode=ro` stopped being parsed as a URI parameter, so the database opened read-write and created a stray file in the parent directory. Worth checking whether the DuckDB path has the same write-side exposure, not just the open failure.

## Fix direction

Percent-encode via `Path.as_uri()` wherever a path becomes a URI, or avoid the URI form entirely and hand DuckDB the path. Lock it with a regression test using a root that contains `?` — the general rule: a path is data, not syntax.

SeoyunL:
A KB path is data. In two places we let it be read as something else.

## The real bug: the KB path was spliced into SQL as syntax

`store/analytics.py` built its `ATTACH` by f-string:

```python
con.execute(f"ATTACH '{db_path}' AS kb (TYPE sqlite, READ_ONLY);")
```

Every character of the path was parsed as SQL. An apostrophe in any ancestor
directory closes the string literal early and the rest of the path is parsed as
code, so `/analytics` dies with a `ParserException` before rendering:

```
"it's a kb"  -> ParserException: Parser Error: syntax error at or near "s"
```

`ATTACH` accepts no prepared parameters, so the path has to reach DuckDB inside
the statement text; doubling `'` (the SQL standard literal escape) is the lever
we actually have. After the fix all of these read fine:

| KB directory | before | after |
|---|---|---|
| `it's a kb` | `ParserException` | OK |
| `semi;dir` | OK | OK |
| `hash#dir` | OK | OK |
| `weird?dir` | OK | OK |

I swept the repo for other path-into-SQL/URI interpolation and this was the only
one: `engine/duckdb_backend.py` runs identifiers through `_quote_ident` (regex
allowlist), `store/db.py`'s `executescript` blocks are static literals, and
`sqlite3.connect`/`duckdb.connect` take the path as an argument, not as SQL.
Closing one hole and leaving the next is the mistake we made in #174; this time
the sweep is in the PR.

## The issue's proposed fix is not implementable

The issue suggests `Path.as_uri()` / percent-encoding to make `?` survive. It
cannot work. The `?` split happens **inside DuckDB's native storage layer**,
below any string we hand it, so there is nothing for us to encode. All six
encoding paths fail (DuckDB 1.5.4, KB root `weird?dir`):

| # | attempt | result |
|---|---|---|
| 1 | raw path | `TransactionException`: cannot open `.../weird.wal?dir/facts.duckdb` |
| 2 | `Path.as_uri()` | `IOException`: cannot open file |
| 3 | manual `?` -> `%3F` | `IOException`: cannot open file |
| 4 | `urllib.parse.quote(path)` | `IOException`: cannot open file |
| 5 | `file://` + quoted path | `IOException`: cannot open file |
| 6 | backslash-escaped `\?` | `IOException`: cannot open file |

Note what #1 does: DuckDB truncates at `?`, appends its `.wal` suffix to the
*head*, and then reports a filename **nobody ever wrote** —
`.../weird.wal?dir/facts.duckdb`.

## So `?` becomes an honest error instead

`?` in a KB root remains an upstream DuckDB limitation. What this PR changes is
how it *reads*. Before: a cryptic WAL error naming an invented path, plus a
half-built 12,288-byte `facts.duckdb` left in the KB. Now the store refuses the
path before opening anything, names the offending character and the path, and
points at the way out:

```
cannot open the DuckDB fact-term store at '/data/weird?dir/facts.duckdb': DuckDB
reads everything after '?' in a database path as connection parameters, not as
part of the filename, so this store can never be opened. Move the KB to a path
with no '?' in it (rename the offending directory, or pass a different KB root)
and retry.
```

Loud is not the same as crashing: the message has to get the user to a recovery.
And no half-built file is left behind — pinned by a before/after directory listing.

The guard is scoped to **DuckDB native storage only**. Analytics' SQLite `ATTACH`
handles `?` perfectly well, so it stays unguarded; over-rejecting would break a
path that works.

## No write-side exposure

Unlike #186's SQLite probe, there is no write path here: `analytics.compute`
attaches `READ_ONLY` and only ever runs aggregate `SELECT`s. The path is never
attacker-supplied either — it is the operator's own KB location. This is a
robustness/correctness fix (a legitimate path breaks a page), not a privilege
boundary.

## Tests

Both new guards are mutation-checked — reverting each fix turns them red:

- revert the `ATTACH` escaping -> 2 failed (`ParserException`)
- drop the `?` fail-fast guard -> 2 failed (leftover `facts.duckdb` assertion bites)

Full suite: 711 passed, 7 skipped. `ruff check` clean. The `?`/`'` tests skip on
Windows, which forbids those characters in filenames.

Closes #188 — the `?` half is closed as *documented upstream limitation surfaced
honestly*, not as a code fix, because the failure matrix above shows no code fix
exists on our side.

### DIFF
--- a/tests/test_analytics.py
+++ b/tests/test_analytics.py
@@ -1,4 +1,7 @@
 # SPDX-License-Identifier: MPL-2.0
+import shutil
+import sys
+
 import pytest
 
 pytest.importorskip("duckdb")  # DuckDB is core; skip only in broken/minimal envs.
@@ -47,3 +50,42 @@ def test_compute_relation_aggregates_use_sqlite_display_mirror(tmp_path):
     s.close()
 
     assert dict(a.by_relation)["has_role"] == 1
+
+
+# A KB path is data, not SQL syntax. ATTACH takes no prepared parameters, so the
+# path is spliced into the statement text and every char below used to be read as
+# syntax: an apostrophe closed the literal and killed /analytics with a
+# ParserException. DuckDB's sqlite reader is happy with '?' (unlike its own
+# native storage, see test_duckdb_fact_terms), so analytics must not reject it.
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Windows forbids these in filenames")
+@pytest.mark.parametrize("dirname", ["it's a kb", "semi;dir", "hash#dir", "weird?dir"])
+def test_compute_reads_a_kb_under_a_path_that_looks_like_sql(tmp_path, dirname):
+    # Seed in a plain directory, then read it from the awkward one, so this test
+    # pins the ATTACH path only and not the native-storage sidecar limits.
+    plain = tmp_path / "plain"
+    plain.mkdir()
+    s = Store(plain / "kb.sqlite")
+    s.init_schema()
+    s.add_fact("A", "is_a", "B", status="confirmed", confidence=0.95)
+    s.close()
+
+    awkward = tmp_path / dirname
+    awkward.mkdir()
+    shutil.copy(plain / "kb.sqlite", awkward / "kb.sqlite")
+
+    a = compute(awkward / "kb.sqlite")
+
+    assert a.available is True
+    assert dict(a.by_status) == {"confirmed": 1}
+
+
+def test_compute_survives_an_apostrophe_that_would_close_the_sql_literal(tmp_path):
+    # The sharp end of the above: an unescaped ' terminates the ATTACH literal
+    # and the rest of the path is parsed as SQL.
+    awkward = tmp_path / "quote'dir"
+    awkward.mkdir()
+    s = Store(awkward / "kb.sqlite")
+    s.init_schema()
+    s.close()
+
+    assert compute(awkward / "kb.sqlite").by_status == []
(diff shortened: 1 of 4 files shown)


justinjoy (approved):
검토 결과 머지를 막을 이슈는 발견하지 못했습니다. 최신 main 기준 병합 결과와 관련/전체 테스트가 통과했습니다.

PR #210, opened 2026-07-13
fix(#198): make the canary see a change it cannot undo, instead of staying silent

### LINKED ISSUE #198: Harden env sandbox canary for unknown pre-run app config entries
## Problem

PR #184 hardens the test environment canary so it snapshots the real app config path before collection and restores regular files, missing entries, and symlink replacements without writing through symlinks. One residual edge case remains: `tests/env_sandbox.py` represents all unreadable or otherwise unknown pre-run entries as `Entry("other")`.

That means an unknown-to-unknown mutation can compare cleanly. For example, if the pre-run `app.json` path is already a directory, and a test changes files inside that directory while leaving the path kind as `other`, `leak_report()` sees `after == before` and returns `None`. The current behavior is intentionally conservative about restore: it refuses to delete or recreate a pre-run unknown entry because the canary did not learn enough to reconstruct it safely. But it also means the canary does not always report that an unknown entry changed.

## Why this matters

The canary now protects the common and dangerous cases: missing config created by a test, real config file overwritten, and symlink replacement. This issue is a follow-up hardening task for unusual pre-existing states at the app config path, such as a directory, special file, or unreadable file.

## Suggested direction

Make `Entry("other")` carry enough metadata to detect unknown-to-unknown changes without trying to restore unreadable contents. Possible options:

- include `lstat()` metadata such as mode, size, mtime/ctime, inode/device where portable enough;
- for directories, capture a shallow listing fingerprint without reading file contents;
- split `other` into more precise kinds, for example `directory`, `unreadable_file`, and `unknown`;
- keep restore conservative, but report when the fingerprint changes and say manual repair is required.

## Acceptance criteria

- A pre-run directory at the app config path whose contents change during the run is reported as a leak.
- A pre-run unknown/unreadable entry is not destructively removed when it cannot be reconstructed.
- Existing cases from PR #184 still pass: missing -> file/directory is removed, file -> file is restored atomically, file -> symlink restores the file without writing through the symlink, and symlink snapshot does not follow the target.

## Context

Follow-up from PR #184 review. This is not blocking PR #184 because it concerns unusual pre-run states that the current implementation deliberately treats as unreconstructable, but it should be tracked as a self-contained hardening issue.

SeoyunL:
The harness had a hole and the canary that was supposed to watch it had several of its own. `POST /settings/root` writes the platform app config resolved from `HOME`, so a web test that forgot to isolate `HOME` rewrote the developer's real `app.json` and repointed their active KB at a temp dir pytest then deleted.

## The sandbox

Three tiers, because a function-scoped `monkeypatch` is too late for most of the leak:

* a **`pytest_configure` seal** — before collection, before any test module's import-time code, before module/session-scoped fixtures;
* an **autouse function fixture** — the per-test sandbox;
* a **`pytest_sessionfinish` canary** — snapshot the real `app.json` at import time, compare at the end, restore and fail loudly if the run touched it.

The isolation tests never request the fixture by name, so deleting the autouse (or the seal) turns them red rather than quietly re-enabling it.

## Review round 1 — the canary could delete the file it exists to protect

**BLOCKER.** `snapshot()` caught *every* `OSError` from `lstat` and called the path `missing`. But `missing` is in `RESTORABLE_KINDS`, and restoring `missing` means `_remove(path)`. So a config directory that was merely unsearchable at import time (EACCES — not empty) produced a `missing` baseline; if the permission came back mid-run, the canary **deleted the user's real `app.json`** and reported it had been "restored to its pre-run state". Had the baseline been a directory, `_remove` would `shutil.rmtree` it.

Now only `ENOENT`/`ENOTDIR` prove absence. Every other errno yields `unknown`, which is outside the whitelist. The whitelist finally means what it advertised: **an entry reaches the destructive path only if we actually read it** — byte for byte (`file`), link for link (`symlink`), or proven absent by errno (`missing`).

Also: the directory fingerprint hashed only each *direct* child's `lstat`, so rewriting `app.json/sub/app.json` — a leaked KB config, the exact shape #198 names — compared clean. It now walks to `_DIR_MAX_DEPTH = 2`.

## Review round 2 — the fix for round 1 opened the same kind of hole

**MAJOR.** The mode-restoring code added in round 1 called **`os.fchmod`, which is Unix-only and does not exist on Windows** — a platform `verinote/config.py` explicitly supports, and which CI (`ubuntu-latest` only) can never catch. Worse, `restore` **deleted the entry and *then* wrote it back**, so on a Windows interpreter the sequence was: delete the user's real `app.json` → `AttributeError` → no restore → `pytest_sessionfinish` dies with an INTERNALERROR instead of saying so. Same category as the round-1 blocker, introduced by its fix.

And it generalised beyond `fchmod`: **any** failure between the delete and the write — `mkstemp` hitting EACCES on a read-only parent, ENOSPC, a chmod raising EPERM — lost the file permanently. `except BaseException` cleaned up the temp file and re-raised; it never put back what had already been deleted.

**Fixed structurally, not symptomatically.** The replacement is staged beside the path and `os.replace`d onto it. `os.replace` overwrites a file or a symlink atomically and in place, so **the path holds the old entry right up until it holds the new one — there is no instant at which it holds neither**, and the delete-then-fail window is gone rather than narrowed. The only delete left on the write path is a *directory* squatting where a file belongs (`os.replace` cannot overwrite one), which by construction is the run's own litter and not the user's entry. The chmod is `hasattr`-guarded and best-effort: a restore that put every byte back must not be undone because the mode would not stick. And `leak_report` no longer raises — a failed restore is **reported**, with the pre-run bytes **parked on disk** for the user, and `pytest_sessionfinish` survives a canary that breaks.

**MAJOR — the fan-out cap was a second silent hole.** `_DIR_MAX_ENTRIES` truncated the listing to the first 256 names plus a *count*, so an in-place edit of the 300th child of a 300-child directory was invisible: the entry is not in the tuple, the count does not move, and editing a file moves no parent's mtime. A `_DIR_MAX_ENTRIES = 1` mutation survived the whole suite. Every child is digested now — the ones past the cap folded into a rolling sha256 — so **the bound costs memory, not sight**.

## What the canary catches, and what it does not

Stated precisely, because an earlier version of this description overclaimed and the next person would have believed it:

* **caught** — creation or deletion down to depth 3 (a create/delete moves its parent's mtime, and that parent is an entry at depth ≤ 2); *modification* at depth 1–2 (rewrite, resize, chmod, re-stamp), each entry's own `lstat` being in the digest; a chmod of the pre-run directory *itself*; a file swapped for a symlink or a directory; a chmod of the real `app.json` with its bytes unchanged; any of the above however **wide** the directory is.
* **not caught** — a modification at depth 4 or deeper, and an in-place rewrite at depth 3 (`app.json/a/b/c.json`): no ancestor's mtime moves and that entry's `lstat` is not in the digest. This is the price of `_DIR_MAX_DEPTH`, it is the **only** detection bound left, and it is **pinned by a test** (`test_the_directory_fingerprint_stops_at_a_documented_depth`, which asserts the constant) so widening it is a deliberate, visible change and the docstring cannot drift from the code.

Detection and repair stay separate powers: an unreconstructable kind (`directory`, `unreadable_file`, `unknown`) is *reported* but never *restored*, since we only ever learned a fingerprint of it.

## Also

* The directory listing is a structured tuple hashed with `hashlib`, not a `,`-joined string: a child named `a:0o100644:0:<mtime>,b` could otherwise forge the digest of a directory holding two children — the separator forgery this repo already fixed in #174.
* `os.geteuid`/`os.mkfifo` are `hasattr`-guarded. A `skipif` condition is evaluated at *import* time, so calling them unguarded broke collection on Windows.
* The whitelist test asserts *behaviour* (every non-restorable kind is refused and the path is left byte-identical), not a restatement of the constant.

## Verification

**939 passed, 7 skipped**; `ruff check .` clean. The developer's real `~/Library/Application Support/verinote/app.json` was verified byte-identical and mode-identical (sha256 `33347a8d…`, 59 bytes, `-rw-r--r--`) before and after the run, with no staging or rescue files left behind.

Mutation — each fix reverted in isolation, one at a time, with the bytecode cache cleared between runs; all are load-bearing:

| # | mutation | result |
|---|---|---|
| M1 | any `lstat` failure folds back into `MISSING` | **RED** — real config would be deleted / rmtree'd |
| M2 | directory fingerprint goes shallow again | **RED** — 3 tests |
| M3 | `st_size` dropped from `_stat_fingerprint` | **RED** — size-only rewrite of an unreadable file |
| M4 | `st_mode` dropped from `_stat_fingerprint` | **RED** — chmod of a FIFO / of the real config |
| M5 | `st_size` dropped from the listing entries | **RED** — size-only change to a child |
| M6 | `restore` stops putting the recorded mode back | **RED** — restore silently re-permissions to 0o600 |
| M7 | **`file`** carries no fingerprint | **RED** — chmod of the real config goes silent |
| M8 | listing collapses back to a `,`-joined string | **RED** — child-name forgery |
| M9 | `restore` deletes before it writes | **RED** — a failed restore left no file on disk |
| M10 | `os.fchmod` called unguarded | **RED** — 2 tests (no-`fchmod` platform / chmod raising EPERM) |
| M11 | `leak_report` lets a failed restore raise | **RED** — the pre-run bytes are never handed back |
| M12 | `pytest_sessionfinish` calls the canary bare | **RED** — the canary takes the session down with it |
| M13 | overflow children truncated to a count again | **RED** — edit past the width bound goes silent |
| M14 | `mtime_ns` dropped from `_stat_fingerprint` | **RED** — mtime-only rewrite of an unreadable file |
| M15 | `st_mode` dropped from the listing entries | **RED** — chmod of a child of a pre-run directory |
| M16 | directory's own `lstat` dropped from `_dir_fingerprint` | **RED** — chmod of the pre-run directory itself |
| M17 | **`symlink`** carries no fingerprint | **RED** — symlink re-created with the same target |
| M18 | errno dropped from a failed listing | **RED** — an unreadable directory digests as an empty one |
| M19 | `_clear_the_way` becomes a no-op | **RED** — 2 tests; the leaked directory is never cleared and the restore cannot land |
| M20 | `_clear_the_way` looks with `stat`, not `lstat` | **RED** — rmtree gets aimed at a symlink into the user's own tree |
| M21 | `_listing` drops its `sorted` | **RED** — an untouched directory fingerprints differently per call (a phantom leak *calls restore*) |
| M22 | `leak_report` catches `Exception`, not `BaseException` | **RED** — a Ctrl-C in `_stage` skips `_park` and the only copy of the config goes with it |
| M23 | `_park` resolves its temp dir lazily instead of at import | **RED** — the rescue copy can land in the sandbox's own soon-to-be-rmtree'd home |

Corrections to the previous description, which this round's review caught: the old M7 row claimed `file`+`symlink` together; mutated **separately**, `file` was red but **`symlink` was green** — M17 is the test that closes it. The suite count read 916 (stale); it is 933. And creations/deletions are caught to depth 3, not 1–2.

M9–M18 were new in round 2; **M19–M23 are new this round**. M3/M4 were red as of round 1.

## Review round 3 — the last delete had no test, and the docstring lied about it

Round 3 found no data-loss bug in the code, but three things worth the round:

* **`_clear_the_way` — the only `rmtree` left in the module — had no coverage at all.** A no-op mutant and an `lstat`→`stat` mutant both survived the whole suite. The code was right; nothing was holding it right. M19/M20 are the tests. Note what M20 pins: with a `stat` there, `rmtree` is *handed* a symlink, and the user's tree survives only because `shutil` happens to refuse that with `ignore_errors=True` swallowing the complaint. That is shutil's invariant, not ours, so the test asserts on the *call*, not just on the survivors.
* **`restore`'s docstring claimed the path is never empty. In one branch it is.** Clearing a leaked directory is a delete before a write, so a `_stage` failure there leaves the path holding nothing — the user's bytes survive on `_park`, not on the ordering. The docstring now says exactly that, and points at the test that pins it; the next reader cannot delete the park on the strength of a promise this function does not make. **This PR exists to stop a canary lying; its own docstring does not get to.**
* **`leak_report` caught `Exception`.** `_stage` re-raises `KeyboardInterrupt` by design — so a Ctrl-C in the window above sailed straight past the rescue net, ending the session with the path empty, `_park` never run, and "check that file by hand" printed about a file that no longer existed. `BaseException` now. (The escape had to be caught explicitly in the test: a `KeyboardInterrupt` does not *fail* a pytest run, it *ends* one — the regression would have read as an absence, not a red.)

Closes #198


SeoyunL:
Thanks — the blocker was real and reproducible, and it inverted the whole point of the PR. All seven are fixed in dd6e321.

**[1] BLOCKER — the canary deleted what it could not read.** You were right that the PR body's central claim was false. `missing` was manufactured from read *failures* and sat inside `RESTORABLE_KINDS`, whose restore branch is `_remove()`. `snapshot` now discriminates on errno: only `ENOENT`/`ENOTDIR` prove absence and yield `MISSING`; every other errno (EACCES, EIO, ELOOP) yields `Entry("unknown", fingerprint=f"lstat_failed:{errno}")`, which is outside the whitelist and therefore unrestorable. The whitelist now actually carries the invariant it claimed: an entry reaches the destructive path only if we read it. Two tests pin it — a pre-run *file* and a pre-run *directory* behind a 0o000 parent, permission restored mid-run, must survive (the directory one is the `shutil.rmtree` blast radius you flagged).

**[2] MAJOR — the shallow listing, and the overclaim.** Fixed and, more importantly, *documented honestly*. `_DIR_MAX_DEPTH = 2` with `_DIR_MAX_ENTRIES = 256` fan-out cap. Both of your failing repros now report: `app.json/sub/app.json` rewritten, and `app.json/sub/x.json` rewritten at the same size. The remaining limit — a modification at depth ≥ 3 — is stated in the `_dir_fingerprint` docstring and **pinned by a test** (`test_the_directory_fingerprint_stops_at_a_documented_depth`), so the doc cannot drift from the code and widening the bound is a deliberate change. The PR body has been rewritten; "catches every realistic shape of the leak" is gone.

**[3] MINOR — M3/M4 survived.** They do not any more. The gap was exactly what you said: no test *detected* an `unreadable_file`/`unknown` change, only `fingerprint is not None`. Added: a size-only rewrite of a pre-run write-only file (mode+mtime pinned), a chmod of a pre-run FIFO (size+mtime pinned), and a size-only change to a child of a pre-run directory. M3 → RED, M4 → RED.

**[4] MINOR — asymmetry, and the silent re-permission.** `file`/`symlink` carry a stat fingerprint now, so a chmod of the real `app.json` is a leak. And `_write_atomically` `fchmod`s the recorded mode back before `os.replace` — `mkstemp` creates at 0o600, so restore was narrowing the user's config permissions. (Worth noting: this half was still unimplemented when I picked the branch up — the test for it was genuinely red, which is how it got caught rather than shipped.)

**[5] MINOR — Windows collection.** `NEEDS_FIFOS = skipif(not hasattr(os, "mkfifo"), ...)` and `NOT_ROOT = skipif(getattr(os, "geteuid", lambda: 1)() == 0, ...)`.

**[6] NIT — change-detector.** Replaced with a parametrized behavioural check: for each of `directory`/`unreadable_file`/`unknown`, `restore()` returns False *and* the path is left byte-identical.

**[7] NIT — separator forgery.** The listing is a structured tuple hashed with `hashlib`; the test builds the `a:0o100644:0:<mtime>,b` forgery and asserts it does not collide.

**Mutation table** (each fix reverted alone, `tests/test_env_isolation.py`, baseline 32 passed):

| # | mutation | result |
|---|---|---|
| M1 | any `lstat` failure → `MISSING` | RED (2) |
| M2 | fingerprint goes shallow | RED (3) |
| M3 | `st_size` out of `_stat_fingerprint` | **RED (1)** — was GREEN |
| M4 | `st_mode` out of `_stat_fingerprint` | **RED (2)** — was GREEN |
| M5 | `st_size` out of listing entries | RED (1) |
| M6 | restore drops the mode chmod | RED (1) |
| M7 | `file`/`symlink` lose the fingerprint | RED (1) |
| M8 | listing back to a `,`-joined string | RED (1) |

Suite: **916 passed, 7 skipped**; `ruff check .` clean. Verified my real `~/Library/Application Support/verinote/app.json` is byte- and mode-identical after every run, including the mutation rounds.

SeoyunL:
All four addressed. The `os.fchmod` finding was the important one — thank you; it was my own round-1 fix reproducing the exact bug it was fixing.

**[1] The delete-then-write window is gone, not guarded.** I did not just wrap `os.fchmod` in a `hasattr`. `restore` no longer deletes anything on the way to a write: the replacement is staged beside the path and `os.replace`d onto it, which overwrites a file *or* a symlink atomically, so the path holds the old entry right up until it holds the new one. There is no longer an instant at which it holds neither — which kills the whole class, not just the `fchmod` instance (ENOSPC, EACCES on a read-only parent, EPERM from the chmod). The only delete left on the write path is a *directory* squatting where a file belongs, which `os.replace` cannot overwrite and which by construction is the run's own litter (the pre-run kind was `file`/`symlink`, so we hold the bytes). The chmod is `hasattr`-guarded *and* best-effort — the bytes are the point, the mode is incidental. `leak_report` no longer raises: a failed restore is reported **and parks the pre-run bytes on disk** with a path to copy back, and `pytest_sessionfinish` no longer dies on it.

Your simulation, re-run against a copy of my real `app.json` on a `del os.fchmod` interpreter:

```
leak_report RAISED: no
does the real app.json still exist? True
bytes restored exactly? True
mode: 0o644
dir now contains: ['app.json']
```

**[2] Went with option A — the hole is gone rather than documented.** Every child is digested now; the ones past `_DIR_MAX_ENTRIES` fold into a rolling sha256 instead of being truncated to a count. So the cap bounds *memory*, not *detection*, and your `_DIR_MAX_ENTRIES = 1` mutation is no longer a hole to survive. `_DIR_MAX_DEPTH` is now the only detection bound left, and it stays pinned + documented as before.

**[3] All six surviving mutations are RED.** Also: while building the harness I hit a false GREEN — `M7'` (drop `file`'s fingerprint) and `M-F` (drop `symlink`'s) delete the *identical* substring, so the mutated source has the same size, and with `st_mtime` at 1-second resolution CPython silently reused the previous mutation's `.pyc`. Cleared the cache between runs; the table below is with `-B` and `tests/__pycache__` removed each time.

| mutation | before | now |
|---|---|---|
| `mtime_ns` dropped from `_stat_fingerprint` | GREEN | **RED** |
| `st_mode` dropped from the listing entries | GREEN | **RED** |
| overflow truncated to a count | GREEN | **RED** (hole removed outright) |
| errno dropped from a failed listing | GREEN | **RED** |
| directory's own stat dropped from `_dir_fingerprint` | GREEN | **RED** |
| `symlink` carries no fingerprint | GREEN | **RED** |
| `restore` deletes before it writes | — | **RED** |
| `os.fchmod` unguarded | — | **RED** (2 tests) |
| `leak_report` lets a restore raise | — | **RED** |
| `pytest_sessionfinish` calls the canary bare | — | **RED** |

11/11 caught. You were right about the M7 row: mutated separately, `file` is red and **`symlink` was green** — the PR table now says `file`, and a symlink-only test closes the other half.

**[4]** Body corrected: 933 passed (was 916), creations/deletions caught to depth 3, M7 row fixed.

Destructive paths re-audited end to end (`_remove`, `_clear_the_way`, `shutil.rmtree`, `unlink`, `os.replace`). My real `~/Library/Application Support/verinote/app.json` is byte- and mode-identical before and after (sha256 `33347a8d…`, 59 bytes, `-rw-r--r--`), with no staging or rescue files left behind.

**933 passed, 7 skipped**; `ruff check .` clean.

SeoyunL:
Round 3 addressed — all five items, `06613f0`.

**[1] MAJOR — `_clear_the_way` had no test.** Two added. The first covers the branch nothing reached (pre-run `file`, post-run non-empty directory → bytes and mode restored, directory gone). The second is the one that kills `lstat`→`stat`, and it needed care: with a `stat` there the user's tree survives *anyway*, because `shutil` refuses an `rmtree` of a symlink and `ignore_errors=True` swallows the complaint. I checked that on the filesystem before writing the test — so a pure survivors-intact assertion **cannot** kill that mutant, and would have been a test that passes for a reason it does not state. The test asserts on the *call*: this module never aims its one `rmtree` at a link. shutil's refusal is shutil's invariant, not ours. `_clear_the_way` also now returns early on `S_ISLNK` explicitly rather than relying on `S_ISDIR` being false for a link.

**[2] MINOR — the docstring lied, and you were right that this is the item that matters.** Confirmed your reading: clearing a leaked directory *is* a delete before a write, so `_stage` failing there leaves the path empty. `restore`'s docstring now says so plainly, names `_park` as what actually carries the bytes in that branch, and points at the test that pins it — so the next reader cannot pull the park out on the strength of a guarantee this function does not make. This PR exists to stop a canary lying; its own docstring does not get an exemption.

**[3] MINOR — `BaseException`.** Fixed, and the test needed a wrinkle worth flagging: a `KeyboardInterrupt` escaping `leak_report` does not *fail* a pytest run, it *ends* one. The first version of the mutation run reported `46 passed` — the session simply stopped. The regression would have read as an absence, not a red. The test now catches the escape and `pytest.fail`s on it, so M22 is a genuine red.

**[4] MINOR — `sorted`.** Pinned: a wide directory listed in reverse-sorted order must fingerprint identically. (Phantom leak → `restore` → a write nobody asked for.)

**[5] NIT — rescue destination.** `_RESCUE_DIR = tempfile.gettempdir()` frozen at import, under the ambient environment, and `_park` writes with `dir=_RESCUE_DIR` — same logic as the REAL paths.

(Your item 6, the `_stage` temp file in the config dir, left alone as instructed.)

**Mutations — the three survivors are dead.** `__pycache__` cleared before every run:

| mutation | before | now |
|---|---|---|
| `_clear_the_way` → `return` (no-op) | 44 passed, **survived** | **RED** — 2 tests |
| `_clear_the_way` `lstat` → `stat` | 44 passed, **survived** | **RED** — `…never_rmtrees_the_directory_a_leaked_symlink_points_at` |
| `_listing` `sorted` → `list` | **survived** | **RED** — `…does_not_depend_on_the_order_the_filesystem_returns` |
| `leak_report` `BaseException` → `Exception` | (new) | **RED** — `…parks_the_bytes_when_the_restore_is_interrupted` |
| `_park` drops `dir=_RESCUE_DIR` | (new) | **RED** — `…parked_outside_anything_the_sandbox_will_delete` |

**Suite: 939 passed, 7 skipped** (was 933; +6 tests). `ruff check .` clean. `tests/test_env_isolation.py`: 50 passed.

**Real `app.json`:** sha256 `33347a8d…e336f1`, 59 bytes, `-rw-r--r--` — identical before and after. A green run leaves **zero** rescue files behind (counted around the run: delta 0). The rescue files sitting in `$TMPDIR` are residue from the *mutant* runs, where the test fails before its cleanup; every one contains fixture bytes (`/real/kb`, `REAL-USER-BYTES`), never anything of the user's.

justinjoy (changes requested):
검토 결과, 현재 변경에는 머지 전에 보완이 필요한 이슈가 있습니다.

1. `before.kind == "missing"` 복구 경로에서 실제 삭제 실패를 성공으로 보고합니다. 예를 들어 pre-run에는 `app.json`이 없었지만 테스트가 파일을 만든 뒤 부모 디렉터리를 unreadable 상태로 바꾸면, `leak_report()`의 after snapshot은 `unknown`이 되고 `restore()`는 `missing` 복구로 `_remove(path)`를 호출합니다. 이때 `_remove()`는 `lstat` 실패를 조용히 반환하므로 파일은 그대로 남는데, `restore()`는 `True`를 반환하고 최종 메시지는 “restored to its pre-run state”라고 보고합니다. 즉 leaked config가 남아 있는데 canary가 복구 성공으로 판단합니다. `missing` 복구 뒤 실제로 경로가 없어졌는지 재확인하거나, `_remove()`가 성공 여부를 반환해 실패 시 `COULD NOT BE RESTORED`로 보고해야 합니다.

   로컬 최소 재현에서 `leak_report()`는 restored 메시지를 반환했지만, `app.json`은 여전히 존재했고 내용도 `b'leaked\n'`였습니다.

2. snapshot equality에 포함되는 metadata를 복구하지 않으면서 “pre-run state”로 복구했다고 보고합니다. `Entry.fingerprint`에는 file/symlink의 `st_mtime_ns`가 들어가지만, `restore()`는 file의 bytes와 mode만 복구합니다. 따라서 `leak_report()`가 “restored to its pre-run state”라고 보고한 직후에도 `snapshot(path) == before`는 false가 됩니다. canary가 감지 기준으로 삼는 상태와 복구 성공 메시지가 서로 다릅니다. 가능한 경우 mtime까지 복구하거나, 메시지/계약을 “bytes and mode restored”처럼 실제 보장 범위로 좁히는 것이 필요합니다.

   로컬 최소 재현에서도 bytes와 mode는 복구됐지만, `mtime_ns`가 달라 `snapshot(path) == before`가 false였습니다.

추가로, pure `env_sandbox` helper 테스트들이 `tests/test_env_isolation.py` 상단의 `pytest.importorskip("fastapi")` 뒤에 있어 FastAPI가 없는 환경에서는 함께 skip됩니다. 이 PR의 핵심은 테스트 harness 자체이므로, 가능하면 FastAPI가 필요 없는 helper-level 테스트는 별도 모듈로 분리하는 편이 더 견고합니다.

확인한 검증 결과:
- `.venv/bin/python -m pytest tests/test_env_isolation.py -q`: 50 passed
- `.venv/bin/python -m ruff check tests/conftest.py tests/env_sandbox.py tests/test_env_isolation.py`: passed
- `.venv/bin/python -m pytest -q`: 939 passed, 7 skipped
- GitHub CI: ruff 및 py3.11/py3.12/py3.13 pytest 모두 통과

SeoyunL:
리뷰 반영했다 — `b01c641` 로 push 됨. 세 항목 모두 처리.

## [1] missing 복구의 거짓 성공 보고 (blocker)

`restore()` 의 missing 브랜치가 `_remove()` 의 조용한 실패 위에서 무조건 True 를 반환하던 것을, **복구 후 `snapshot(path) == MISSING` 재확인**으로 교체 — 실제로 경로가 사라졌을 때만 성공이고, 아니면 `COULD NOT BE RESTORED` 로 보고한다. 리뷰의 재현 시나리오(pre-run 부재 → 테스트가 파일 생성 → 부모 디렉터리 0o000)를 그대로 테스트로 고정: leaked 파일이 남으면 복구 실패로 보고돼야 red/green 이 갈린다 (teardown 에서 권한 원복).

## [2] "restored to its pre-run state" 과잉 주장

옵션 (a) — **mtime 복구**를 채택했다. 이 모듈의 존재 이유가 '카나리아가 거짓말하지 않기'이므로 메시지를 좁히기보다 주장을 참으로 만드는 쪽이 맞다. `_restore_mtime()` 이 `os.utime(ns=...)` 로 ns 단위 복구(symlink 는 `follow_symlinks=False`, lutimes 없는 플랫폼만 문서화된 예외로 좁힘). 이제 복구 성공 직후 `snapshot(path) == before` 가 **실제로 참**이며, 그걸 그대로 단언하는 테스트를 추가했다.

## [3] 헬퍼 테스트의 fastapi 게이트 (권고 채택)

순수 헬퍼 테스트를 `tests/test_env_sandbox_helpers.py` 로 분리 (fastapi 불필요). 테스트 이름을 그대로 옮겨 기존 뮤테이션 표와 1:1 대응 유지. `test_env_isolation.py` 의 fastapi 게이트는 실제로 웹 스택을 쓰는 테스트 1개로 축소. fastapi 차단 시뮬레이션으로 FastAPI 없는 환경에서도 하네스 자체 테스트가 수집·실행됨을 확인.

## 검증

- 전체 스위트 **942 passed, 7 skipped** · ruff clean
- 뮤테이션: [1] 재확인을 `return True` 로 되돌림 → red (거짓 'restored' 메시지 검출), [2] `_restore_mtime` 호출 제거 → red 2건 (fingerprint 의 mtime_ns 불일치)

PR #304, opened 2026-07-19
refactor(#292): retire Store.set_status(), the unguarded status write

### LINKED ISSUE #292: refactor(store): retire `set_status()` — a public blind write that bypasses every transition guard
## 배경

PR #263 (#231/#232/#257) 에서 갈라져 나온 후속이다.

그 PR 은 리뷰 게이트의 전이 경로를 전부 원자화하고 **"사람의 거부(superseded)는 종착 상태이고 어떤 프로덕션 경로로도 빠져나갈 수 없다"** 는 불변식을 세웠다.

| 경로 | superseded 탈출 |
|---|---|
| `toggle_review` | 불가 — superseded 조기 반환 + conditional write |
| `accept_fact` | 불가 — tier check + conditional write |
| `reject_fact` | 불가 — superseded 만 씀 |
| `auto_accept_fact` | 불가 — tier 재확인 + conditional write |
| `accept_review_facts_for_source` | 불가 — SELECT 가 REVIEW_STATUSES 필터, 한 트랜잭션 안 |
| `amend_fact` | 불가 — status 무접촉 |
| **`set_status`** | **가능 — blind write** |

## 문제

`set_status()` 는 `Store` 의 공개 메서드인데 어떤 상태로든 **무조건** 쓴다. 이름이 무해해 보이는 만큼, 앞으로 전이 경로를 추가하는 사람이 위 가드를 전부 우회하는 **가장 쉬운 길**이다.

지금은 `superseded` 를 되살릴 수 있는 유일한 남은 문이다.

## 왜 지금이 싼가

PR #263 의 수정들로 **프로덕션 호출부가 0개가 됐다** (`acceptance.py` 가 마지막 호출부였다). 현재 호출부는 테스트 픽스처뿐이다.

## 왜 PR #263 에서 안 했나

1. 테스트의 "fact 를 상태 X 로 놓고 시작한다" 는 요구 자체는 정당하다. 그건 전이 API 가 아니라 **픽스처용 seam** 이 필요한 별개 설계 문제이고, 전이 검증 PR 안에서 즉흥적으로 결정할 일이 아니다.
2. 호출부가 `test_web.py` / `test_store_fact_terms.py` / `test_verify.py` / `test_acceptance.py` 에 흩어져 있는데 일부는 열린 PR (#209/#229/#236) 이 점유 중이라 지금 건드리면 범위 경계를 깬다.

## 할 일 (제안)

픽스처가 쓸 명시적 seam 을 정하고 (예: 이름이 의도를 드러내는 테스트 전용 헬퍼, 또는 삽입 시점에 상태를 받는 경로), `set_status()` 를 공개 API 에서 걷어낸다. 열린 PR 들이 정리된 뒤에 하는 게 안전하다.

## 참고

- PR #263 — https://github.com/semantic-reasoning/verinote/pull/263

SeoyunL:
Closes #292

## What

`Store.set_status()` is gone. It wrote any status onto any fact with no tier
check, which made it the shortest way around every transition guard in the
review gate — including the rule that a human's rejection is terminal.

Deleting it outright rather than deprecating it: there were **no production
callers** (by static Python reference; `verinote/**` held the definition and
nothing else), it was not in `verinote/store/__init__.py`'s `__all__`, and this
repo publishes no external API contract. A shim would have left the bypass in
place under a new name.

The six remaining callers were all tests, and each wanted a *transition* rather
than a starting state, so each moved to the real guarded route
(`accept_fact` / `reject_fact`).

## No new test-only seam

The issue asked for a fixture seam before removal. One already exists:
`add_fact(status=...)` is the insertion seam that picks a fact's *starting*
status, and that is what tests reaching for `set_status` actually needed in the
insert case. Adding a test-only blind-write helper would have moved the hole
the issue points at, not closed it.

## What the substitution surfaced

`test_provenance_renders_trust_dossier_sections` was seeding `confirmed` and
then transitioning to `accepted` — a transition **production cannot perform**,
since `confirmed` is outside `REVIEW_STATUSES`. The test's premise was wrong,
not the guard: its seed is now `needs_review`, an actually reachable history.
Restore the old seed and `accept_fact` is refused by the guard, writes nothing,
logs no `accepted` event, and the timeline assertion goes red.

## The guard, and how it got here

The regression guard went through four rounds. Each round is a separate commit,
because what each one *failed* to catch is the useful part of the record.

1. **Freezing `Store`'s public surface** (67 names, partitioned into transition
   and non-transition). It looked strict and was not: attaching a real blind
   write and adding its name to the non-transition list left all guards
   **green**, while adding an unrelated `source_count()` went **red** at once.
   It detected novelty of names, not blind writes — the wrong thing, in both
   directions.
2. **Scanning Store's own source** for `UPDATE facts SET ... status`. The
   pattern stopped at the first quote, so `datetime('now')` hid anything
   written after it: `SET updated_at = datetime('now'), status = ?` scanned
   **clean**. Column order alone defeated it.
3. **That fix was itself unguarded.** Reverting the pattern kept the whole suite
   green, because today's five writers all happen to name `status` first. The
   detection rule is now pinned by its own tests, with a negative control so it
   has to discriminate rather than answer yes to every `UPDATE` on `facts`.
4. **`<=` → `==`.** A subset assertion is satisfied by the empty set, so
   detection collapsing to nothing would have passed silently. Equality makes
   that a failure.

## What this guard does not cover

Detection reads **one method body at a time**. A blind write factored into a
private helper, with the public method delegating to it, is invisible to it:

```python
def _blind_status(self, fact_id, status): ...   # SQL lives here
def force_status(self, fact_id, status):        # public, unguarded
    with self._lock:
        self._blind_status(fact_id, status)
```

Verified, not theorised: the scan stays green while
`force_status(fact_id, "confirmed")` brings a superseded fact back. That state
is reached by an ordinary refactor — extracting shared SQL — rather than by
anyone trying to evade the guard. The comments and module docstring claim only
the inlined case.

Put as the question that found it: **keeping this guard green while introducing
a blind write costs zero lines of allowlist edit — just moving the SQL down one
frame into a private helper.** Raising the detection unit from the function body
to the module (ast) is follow-up work, deliberately not attempted here; the
guard has been rewritten enough times in one PR.

## Verification

- `pytest tests -q` → **1233 passed** (1229 on main; +4 guards)
- `ruff check .` clean
- Mutation evidence, each confirmed red by hand: dropping `accept_fact`'s
  `REVIEW_STATUSES` early return (`verinote/store/db.py:1296`) breaks the sweep;
  restoring `set_status()` breaks the scan; a column-order-swapped blind write
  breaks the scan; reverting the pattern breaks the rule's own test; collapsing
  detection breaks the equality assert.

## Follow-ups (not opened here)

1. Raise the detection unit to an ast-based module scan so private helpers are
   in view.
2. Encode the transition table as a SQLite trigger, so the invariant lives in
   the schema instead of in a test.
3. Decide the policy for `amend_fact` rewriting the content of a superseded
   (rejected) fact — existing behaviour, currently unspecified.

### DIFF
--- a/tests/test_acceptance.py
+++ b/tests/test_acceptance.py
@@ -258,7 +258,7 @@ def test_accept_recommendation_rejects_previous_rejection(tmp_path):
         source_id=source_id,
         job_id=job_id,
     )
-    s.set_status(fact_id, "superseded", action="rejected")
+    s.reject_fact(fact_id)
 
     recommendation = accept_recommendation(s, fact_id)
 
--- a/tests/test_review_gate.py
+++ b/tests/test_review_gate.py
@@ -846,7 +846,12 @@ FACT_TRANSITION_METHODS = frozenset(
     }
 )
 
-_STATUS_WRITE = re.compile(r"UPDATE\s+facts\s+SET[^\"']*status", re.IGNORECASE)
+# The assignment list of every `UPDATE facts SET ... WHERE`. Matching the whole
+# SET clause rather than stopping at the first quote is what makes column order
+# irrelevant: `datetime('now')` sits between `SET` and `status` the moment
+# someone writes `updated_at` first, and a quote-sensitive pattern would read
+# that as "no status write" and wave the row straight past both guards.
+_STATUS_WRITE = re.compile(r"UPDATE\s+facts\s+SET(.*?)WHERE", re.IGNORECASE)
 
 
 def _public_methods(cls) -> frozenset[str]:
@@ -857,11 +862,14 @@ def _public_methods(cls) -> frozenset[str]:
     )
 
 
+def _writes_status(method) -> bool:
+    source = " ".join(inspect.getsource(method).split())
+    return any("status" in clause for clause in _STATUS_WRITE.findall(source))
+
+
 def _status_writers(cls) -> frozenset[str]:
     return frozenset(
-        name
-        for name in _public_methods(cls)
-        if _STATUS_WRITE.search(" ".join(inspect.getsource(getattr(cls, name)).split()))
+        name for name in _public_methods(cls) if _writes_status(getattr(cls, name))
     )
 
 
--- a/tests/test_review_gate.py
+++ b/tests/test_review_gate.py
@@ -6,6 +6,9 @@ nor an accept may revive a `superseded` fact, and the status-changing web routes
 apply auto-accept so a decision on one fact reveals promotions of its siblings.
 """
 
+import inspect
+import re
+
 import pytest
 
 pytest.importorskip("fastapi")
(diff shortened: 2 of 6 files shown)


justinjoy (changes requested):
차단 이슈가 있습니다.

`Store.set_status()` 제거 방향은 맞지만, PR이 새로 추가한 sweep 대상 안에 아직 같은 등급의 unguarded status write가 남아 있습니다. `accept_review_facts_for_source()`는 먼저 review-tier row를 조회한 뒤 각 row를 `UPDATE facts SET status = confirmed ... WHERE id = ?`로 갱신합니다. 반면 `accept_fact()`, `reject_fact()`, `toggle_review()`, `auto_accept_fact()`는 모두 관측한 status를 `WHERE id = ? AND status = ?`에 넣어 stale transition을 방지합니다.

이 차이 때문에 다른 Store connection에서 조회 이후 같은 fact를 reject/supersede한 경우, bulk accept 경로는 같은 “관측한 상태에 대한 조건부 전환” 계약을 지키지 않습니다. 현재 `test_no_public_transition_revives_a_superseded_fact`는 호출 전에 이미 superseded인 경우만 검증하므로 이 중간 전환 경합을 잡지 못합니다.

`accept_review_facts_for_source()`도 각 row의 observed status를 조건으로 업데이트하고, `rowcount == 0`이면 로그와 accepted 결과에 포함하지 않도록 맞춰 주세요. 이 경합 또는 최소한 stale row를 덮어쓰지 않는 동작에 대한 회귀 테스트도 필요합니다.

PR #305, opened 2026-07-19
fix(#298): derive the query intent field lists from the schema

SeoyunL:
Refs #298

`query_intent.py` kept two hand-written copies of `QUERY_INTENT_SCHEMA`'s property
names. This derives them, and makes the half-wired states that derivation exposes
fail loudly instead of quietly.

## Closed

**The name lists.** `_NULLABLE_STRING_FIELDS`, `QUERY_INTENT_COMPARISON_DOMAINS`,
`QUERY_INTENT_BLANK_NULLABLE_FIELDS` and `QUERY_INTENT_FIELDS` are now derived
from the schema. The derivations take the schema as an argument rather than
reading the module constant, which is what makes them testable against a
synthetic schema: an inline comprehension over the real constant has the same
extension as the hand list it replaced, so nothing could tell the two apart.

The predicate is `set(type) == {"string", "null"}`, not `"null" in type` — the
looser form also matches `relation_candidates` (`["array", "null"]`) and the
three target properties (`["object", "null"]`), whose values are not strings and
must not be trimmed or blank-normalised as though they were.

**The silent drop.** Deriving the allow-list alone moved the failure rather than
closing it. The `QueryIntent` construction names its kwargs by hand, so a
property added to the schema stopped being refused as an unexpected field and
started being *accepted and discarded*: a blank or off-enum value came back as
`None` instead of being rejected. That is quieter and worse than the error it
replaced. The module now refuses to import while it admits a name it never reads.

**Diagnosis attribution.** That failure is raised at import, outside the parse
path entirely, and with a type `parse_query_intent`'s
`except (KeyError, TypeError, ValueError)` does not convert. Both halves of the
comparison are module constants, so no provider output can reach it; reporting it
as `LLMError: query intent output did not match schema` would pin a local wiring
bug on the provider and send the reader after the wrong thing.

## Not closed — why this is `Refs` and not `Closes`

The issue's acceptance is that adding a property to the schema makes it
blank-rejecting **with no further action**. That is not what this branch
delivers. The parser kwarg and the dataclass field are still wired by hand:

```
schema + dataclass field only, parser not wired
  -> RuntimeError: query intent schema has properties the parser never reads: unit;
     add them to the QueryIntent construction in _parse_query_intent_object

after wiring the kwarg
  blank  -> LLMError: unit must be one of kg, m, got ''
  'lb'   -> LLMError: unit must be one of kg, m, got 'lb'
  'kg'   -> ACCEPTED, unit='kg'
```

So what changed is silent-discard to loud-failure, not manual to automatic.
Completing the acceptance means deriving the construction itself, which is left
as follow-up (see below).

## Known trap

`_PARSED_INTENT_FIELDS` — the list of names the parser actually reads — is held
against the *schema*, not against the kwargs it mirrors. Adding a name to it
without wiring the corresponding kwarg brings the silent drop straight back, with
the guard quiet and the suite green (reproduced with a `locale` property during
review). Reaching that state means ignoring an error message that names the fix,
so it is narrow rather than likely, but it is real and the code comments say so.

## Non-vacuity

Every guard here was checked by reverting the fix and confirming the test goes
red, not merely by watching it pass.

| Mutation | Result |
| --- | --- |
| M1 derivation → the four-name hand list | 3 failed |
| M2 predicate → `"null" in type` | 21 failed |
| M3 drop `reason` from the dataclass | 1 failed |
| M4 allow-list → the nine-name hand list | 4 failed |
| M6 delete the import-time guard | 1 failed |
| M7 derivation → an extensionally identical nine-name literal | 1 failed |
| M8 guard hard-wired to always raise | package unimportable |

**M7 is the load-bearing one.** Replacing the derivation with a literal that has
exactly today's value still goes red, which is what distinguishes a real
derivation from a hand list that happens to agree. An earlier version of this
test asserted `"unit" in _intent_field_names(synthetic)` — a property of the
derivation rather than of the parser — and passed while the silent drop was live;
the current test builds the module against a synthetic schema and so exercises
what the derivation is for. M8 is why there is a second test pinning that the
guard still passes on the real schema: a guard wired to always raise would
otherwise satisfy the first one while making the package unimportable.

The tests load an isolated module copy via `spec_from_file_location` rather than
`importlib.reload`; a failed reload leaves the real module half-rebuilt, and even
a successful one swaps the identity of `QueryIntent`/`IntentTarget` out from under
every test that imported them, breaking dataclass equality later in the session.

## Blast radius

If the guard fires, `verinote.pipeline` and anything importing through it fail to
load (`import verinote` itself still works). The schema is consumed only at import
(`schema.py:131`) and the four adapters only read it, so there is no path by which
it changes at runtime — this is a development-time failure only. The traceback is
four frames and its last line names the missing property and the function to fix.

## Verification

`1236 passed`, `ruff check` clean.

## Follow-up

Derive the construction itself so `_PARSED_INTENT_FIELDS` can go away: a
field-to-parser dispatch table (`_INTENT_FIELD_PARSERS`) would make the list *be*
the construction rather than a claim about it, removing both the trap above and
the manual wiring step. That is what would complete the issue's acceptance.

### DIFF
--- a/tests/test_llm_schema.py
+++ b/tests/test_llm_schema.py
@@ -1,4 +1,7 @@
 # SPDX-License-Identifier: MPL-2.0
+import copy
+import dataclasses
+
 import pytest
 
 from verinote.llm import LLMError
@@ -12,9 +15,27 @@ from verinote.pipeline.query_intent import (
     QUERY_INTENT_BLANK_NULLABLE_FIELDS,
     QUERY_INTENT_COMPARISON_DOMAINS,
     IntentTarget,
+    QueryIntent,
     QueryIntentKind,
+    _blank_nullable_fields,
+    _comparison_domains,
+    _nullable_string_fields,
     parse_query_intent,
 )
+
+
+def _synthetic_intent_schema():
+    """QUERY_INTENT_SCHEMA plus two nullable string properties it has never had.
+
+    A hand-written name list and a schema derivation agree on today's contract, so
+    only a schema the module has never seen tells them apart: the derivation
+    answers with `unit` and `note`, the list cannot.
+    """
+    schema = copy.deepcopy(QUERY_INTENT_SCHEMA)
+    schema["properties"]["unit"] = {"type": ["string", "null"], "enum": ["kg", "m", None]}
+    schema["properties"]["note"] = {"type": ["string", "null"], "minLength": 1}
+    schema["required"] = list(schema["properties"])
+    return schema
 
 
 def _intent_payload(**overrides):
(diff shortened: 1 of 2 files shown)


justinjoy (approved):
검토 결과 blocking issue는 없습니다.

`QUERY_INTENT_SCHEMA`에서 nullable string field, enum domain, blank-nullable field, parser allow-list를 파생하도록 바뀐 점과, schema property가 parser construction에 연결되지 않은 경우 import-time에 실패하도록 한 guard를 확인했습니다. 현재 schema 기준 동작 보존과 drift 방지 테스트가 충분하며 GitHub CI도 모두 통과했습니다.

PR #313, opened 2026-07-20
fix(#294): move the trust-signal group label onto a role that may carry it

### LINKED ISSUE #294: a11y: trust-labels 의 aria-label 이 name-prohibited div 에 걸려 있어 보조기술에 안 읽힐 수 있다
## 배경

PR #284 (#278) 리뷰 중 발견됐다. **이 PR 이전부터 있던 것**(`b0ef0cc "Expand provenance into trust dossier"`)이라 그 PR 범위 밖으로 뒀다.

## 문제

`verinote/web/templates/provenance.html:11`:

```html
<div class="trust-labels" aria-label="trust signals">
```

`<div>` 의 implicit ARIA role 은 `generic` 이고, 이 role 은 author-provided accessible name(`aria-label`/`aria-labelledby`)을 **허용하지 않는다**(name-prohibited). 표준을 따르는 보조기술은 이 라벨을 accessible name 으로 쓰지 않을 수 있다.

즉 DOM 에는 `aria-label="trust signals"` 가 보이지만 스크린리더 출력에는 안 나올 수 있다 — **고쳤다고 믿는데 실제로는 조용히 아무 일도 안 하는** 상태다.

## 이것이 #278 과 같은 계열인 이유

#278/PR #284 가 정확히 같은 문제를 term/string 슬롯에서 고쳤다. 거기서 채택한 해법:

- `aria-label` 을 버리고 **실제 텍스트 노드**로 노출: `<span class="visually-hidden">…</span>`
- 헬퍼 CSS 는 clip-path 관용 구현 (`app.css` 에 이미 추가됨 — **재사용 가능**)
- 테스트는 markup substring 이 아니라 `HTMLParser` 로 announced/visible 텍스트를 분리해 검증하고, `span[aria-label]` 같은 name-prohibited 패턴을 **금지 핀**으로 잠금

## 할 일 (제안)

`trust-labels` 도 같은 방식으로 옮긴다. PR #284 가 머지된 뒤 하면 헬퍼 CSS 와 테스트 유틸리티를 그대로 재사용할 수 있다.

주의: 여기서 노출하려는 게 **그룹 라벨**(개별 값이 아니라 "이 묶음이 trust signals 다")이므로, term/string 슬롯과 달리 값+kind 병기가 아니라 **의미 있는 role/element 로 감싸는 쪽**(예: naming 이 허용되는 role, 또는 제목 요소)이 더 맞을 수 있다. #284 의 term 슬롯 해법을 기계적으로 복사하지 말고 이 자리에 맞는 걸 판단할 것.

또한 PR #284 의 `span[aria-label]` 금지 핀을 **`div[aria-label]` 까지 넓힐지** 함께 검토한다 — 넓히면 이 결함이 재발할 수 없다.

## 참고

- PR #284 — https://github.com/semantic-reasoning/verinote/pull/284 (term/string 슬롯에서 같은 문제를 해결, 헬퍼 CSS·테스트 패턴 제공)
- #278, #222, #276

SeoyunL:
Closes #294

## The defect

`provenance.html:11` hung the group label on a host that is not allowed to carry one:

```html
<div class="trust-labels" aria-label="trust signals">
```

A `<div>` has the implicit ARIA role `generic`, and `generic` is **Name From: prohibited** — a conforming AT is free to ignore an author-provided `aria-label` on it. So the label was not merely redundant, it was droppable: the badges could be announced as a run of loose words with nothing saying what they were signals *of*.

The fix gives the label a host that may hold one, rather than moving the label:

```html
<div class="trust-labels" role="list" aria-label="trust signals">
  <span class="badge trust-…" role="listitem">…</span>
```

`role="listitem"` on the badges is not decoration — `list` has **required owned elements**, so the two go together or the container announces a list that owns nothing.

**No `app.css` change.** `role` attributes do not affect rendering, so `.trust-labels`' flex layout and `.badge` styling are untouched. Changed files are `provenance.html` plus one new test file.

**Why not `<ul>/<li>`?** WebKit strips list semantics from a list styled `list-style: none`, so the accessible-name problem would come back and an explicit `role="list"` would have to be re-added anyway — on top of needing list-marker resets in `app.css`.

**Scope.** A global search found `aria-label` in 10 places across the templates; the other 9 sit on hosts where naming is permitted (`<nav>`, `<select>`, `<input>`). Within that grep's range, this was the only one on a name-prohibited host.

## The guards were breached twice, and that is the point

The regression guards in `tests/test_trust_labels_a11y.py` were written first and then attacked. Two holes turned up, both worth recording:

1. **A conditional `pytest.skip` hid a real defect.** The owned-elements guard skipped itself when the container was not a `list`. Swap the role to `region` and keep the `role="listitem"` badges and the suite reported *3 passed, 1 skipped* — while the markup contained **orphan listitems**, items owned by no list, which makes an AT report position and count against something that does not exist. Guard 1 waves that mutant through because `region` does permit a name. Fixed by dropping the skip for an assertion in both directions: if the container is a `list`, every child is a `listitem`; if it is not, no child may be one. The file now has **zero skips**.

2. **The name-prohibited pin only checked that a `role` attribute was *present*.** So `<div role="presentation" aria-label="sneaky">` cleared it — *4 passed* — despite being exactly as unannounceable as a bare div. Fixed with a `NAME_PROHIBITED_ROLES` constant (`generic`, `presentation`, `none`, `paragraph`) checked against the role's value.

The pin is deliberately stated as "a div/span whose role prohibits a name" rather than "any `div[aria-label]`". The latter would fail on this very fix, which is a legitimately named div.

## Mutation testing

| Mutant | Result |
|---|---|
| M1 — remove `role="list"` | 3 failed |
| M2 — remove `aria-label` | 1 failed |
| M3 — remove `role="listitem"` | 1 failed |
| M4 — `role="region"`, keep listitems | 1 failed |
| M5 — add `<div role="presentation" aria-label="sneaky">` | 1 failed |

All five are red, and under each one **every other test in the suite still passes** — the new guards are what catch them, with nothing riding along.

M1–M3 are deletions of the prescribed markup; M4–M5 are substitutions. That distinction mattered here: a guard exercised only against deletion checks that the prescription is still present, and stays blind to a path that swaps it for something else. Both holes above were substitutions.

## Verification

- `PYTHONPATH=$PWD python -m pytest tests -q` → **1233 passed**, **zero skips** (confirmed with `-rs`)
- `ruff check .` → clean

## Follow-ups (not in this PR, no issues filed)

- `fact_row.html:19-40` renders the same badge bunch with **no group name at all** — same family of defect, separate change.
- Widening the pin's coverage from these 2 render paths to every template.
- The remaining name-prohibited roles in the spec (`caption`, `code`, `strong`, …), none of which the current templates use.

### DIFF
--- a/tests/test_trust_labels_a11y.py
+++ b/tests/test_trust_labels_a11y.py
@@ -1,0 +1,207 @@
+# SPDX-License-Identifier: MPL-2.0
+"""The trust-signal badges need a *group* label that assistive tech can reach (#294).
+
+The bunch used to sit in `<div class="trust-labels" aria-label="trust signals">`. A
+plain `<div>` has the implicit ARIA role ``generic``, and ``generic`` is
+*name-prohibited*: a conforming AT may drop an author-provided ``aria-label`` on it
+entirely, so the only thing announced was a run of loose badge words with nothing
+saying what they were signals *of*.
+
+This is a different contract from the triple slots in ``test_term_kind_a11y.py``,
+which is why it lives in its own file. There the label belongs to a *value* and the
+fix is to put the kind into a real text node; here the label names a *group*, and
+moving it into the visible text would just be noise on the page. The fix instead
+gives the label a host that is allowed to carry one -- ``role="list"`` -- and, because
+``list`` has required owned elements, marks each badge ``role="listitem"``.
+
+Parsing is stdlib ``html.parser`` on purpose; see the docstring of
+``test_term_kind_a11y.py`` for why an undeclared transitive ``lxml`` is not used.
+The assertions below deliberately go through the parsed DOM rather than substring
+checks on the markup: ``'aria-label="trust signals"' in html`` would still pass with
+the role stripped back off, i.e. it would pass on the bug this file exists to pin.
+"""
+
+from __future__ import annotations
+
+from html.parser import HTMLParser
+
+import pytest
+
+pytest.importorskip("fastapi")
+from fastapi.testclient import TestClient  # noqa: E402
+
+from verinote.config import Config  # noqa: E402
+from verinote.store.fact_input import structural_term  # noqa: E402
+from verinote.web import create_app  # noqa: E402
+
+GROUP_CLASS = "trust-labels"
+
+# Roles that permit an author-provided accessible name and suit a bunch of labels.
+# ``generic`` (the implicit role of a bare div/span) is pointedly not among them.
+NAMING_ROLES = {"list", "group", "region"}
+
+# Tags whose implicit role is ``generic``. Naming them requires an explicit role.
+GENERIC_TAGS = {"div", "span"}
+
+NAME_ATTRS = ("aria-label", "aria-labelledby")
+
+# HTML void elements never get an end tag, so they must not enter the open-tag stack.
+VOID = {
+    "area", "base", "br", "col", "embed", "hr", "img", "input",
+    "link", "meta", "param", "source", "track", "wbr",
+}  # fmt: skip
+
+
+class _Markup(HTMLParser):
+    """Pull out the trust-label group, its direct children, and every named generic.
+
+    ``group`` is the attributes of the ``.trust-labels`` element, ``children`` the
+    attributes of each *direct* child element of it (nested descendants do not count
+    -- required owned elements are about the immediate children). ``named_generics``
+    collects every ``div``/``span`` that carries an accessible name without declaring
+    a role, which is exactly the name-prohibited pattern.
+    """
+
+    def __init__(self) -> None:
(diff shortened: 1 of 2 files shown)


justinjoy (approved):
검토 결과 blocking issue는 없습니다.

`provenance.html`의 trust label 묶음에 `role="list"`와 child `role="listitem"`를 함께 부여해 aria-label이 name-prohibited generic role에 매달리는 문제를 해소한 것을 확인했습니다. 새 parser 기반 a11y 테스트가 label host, required owned elements, name-prohibited generic role 회귀를 직접 검증하고 있으며 CI와 focused 테스트가 통과했습니다.

PR #314, opened 2026-07-20
fix(#221): tell a rule's auto-accept apart from a person's confirm in the badge

### LINKED ISSUE #221: UI: a rule-accepted fact is indistinguishable from a human-confirmed one
## 문제

`verinote/web/static/app.css:36`:

```css
.badge-confirmed, .badge-accepted { color: var(--ok); border-color: var(--ok); }
```

두 상태가 **한 규칙으로 묶여 있어 화면에서 똑같이 보인다.** 그런데 이 둘은 전혀 다른 사건이다:

| 상태 | 누가 만들었나 | 코드 |
|---|---|---|
| `confirmed` | **사람**이 Accept 버튼을 눌렀다 | `web/app.py:924` — `set_status(id, "confirmed", action="accepted")`, actor 기본값 `human` (`db.py:1066`) |
| `accepted` | **규칙**이 자동 승인했다 | `pipeline/acceptance.py:67-72` — `actor="rule"`, `rule_name="corroborated_no_conflict"` |

`auto_accept_recommendations` (#91 에서 도입, 기본 off)를 켜면 사람 클릭 없이 팩트가 엔진 입력이 된다. 감사 장부를 파는 제품에서 **"누가 이걸 통과시켰는가"가 화면에서 지워지는 것**은 그냥 스타일 문제가 아니다. 리뷰 행은 actor 를 아예 렌더하지 않는다(`partials/fact_row.html` 에 actor/rule_name 참조 없음). 사용자가 볼 수 있는 단서는 `confirmed` / `accepted` 라는 **단어 두 개뿐**이고, 그마저 같은 색·같은 모양이다.

(actor 자체는 provenance 페이지에는 나온다 — `provenance.html:115-122`. 즉 데이터는 있는데 판단 화면에서만 안 보인다.)

## 고칠 방향

- `.badge-accepted` 를 `.badge-confirmed` 에서 분리한다. 같은 ok 톤이되 **글리프를 달리한다**: 사람 승인 `✓`, 규칙 승인 `✓` + 작은 기어(⚙) 첨자.
- 스타일은 **상태 문자열에서만** 파생한다. 하드코딩 클래스나 데모 데이터 지름길 금지.

## 검증

`auto_accept_recommendations` 를 켜고 추출 → 규칙이 승인한 팩트와 사람이 승인한 팩트가 리뷰 큐에서 **다르게** 보일 것.

관련: #91 (자동수용 도입).

SeoyunL:
Closes #221

## What

`confirmed` (a person clicked Accept) and `accepted` (a rule auto-promoted the fact) were styled by a single rule — `.badge-confirmed, .badge-accepted { color: var(--ok); border-color: var(--ok); }` — so the review screen rendered the two identically. The reviewer could not tell which had happened.

`.badge-accepted` is now split out and carries two non-colour channels: `border-style: dashed` and a `::before { content: "⚙" / "" }` glyph. Both badges keep `var(--ok)`. No new tokens, no template change — `badge-{{ f['status'] }}` already emits the two classes separately, so the styling stays derived from the status string.

## Why the two states differ

`confirmed` is written by `accept_fact()` in `verinote/store/db.py`, whose audit-log actor defaults to the human user. `accepted` is written by `auto_accept_fact()` with `actor="rule"` and a `rule_name`. Both are in `ENGINE_STATUSES`, so a rule-promoted fact reasons with exactly the authority of a human-accepted one — which is what makes the visual collapse worth fixing rather than cosmetic.

## Why `.badge-confirmed` gets no glyph

That class is reused as a generic ok chip: `dashboard.html:52` renders "covered" with it and `report.html:12` renders "OK". Stamping an approval mark there would put an accept glyph on text that records no accept decision. The symmetric `✓` the issue sketches belongs with #226's global glyph mapping.

## Not colour alone

#226 is open on verdicts that depend on colour alone, so the load-bearing guard asserts that **the two declaration sets still differ once every colour property is stripped**. A fix that only re-tints one badge is red.

## The guard was defeated twice before it held

This is the part worth reading.

1. **An inert declaration plus a blank glyph passed.** `font-style: normal` (the inherited default, so no visual effect) together with `content: "" / ""` gave **6 passed while the two badges were pixel-identical**. The cause was `assert content` treating the string `'"" / ""'` as truthy — "the property is set" is not "a glyph is drawn". Fixed by splitting the `/ ""` alt half off and asserting the drawn half is non-empty.

2. **A colour hidden in a shorthand slipped through.** `border: 1px solid var(--muted)` changes nothing but the hue, yet was counted as a non-colour difference because stripping only looked at `color` / `border-color` / `background`. Fixed by stripping the shorthands too. Deliberately **not** via `startswith("border")` — that would strip `border-style` as well, gutting the very channel the guard protects.

## Mutation evidence

Each mutant was applied to a `git archive` export, never the worktree.

| Mutant | Change | Result |
|---|---|---|
| A | revert to the one grouped rule, glyph deleted | 4 failed |
| B | difference made colour-only | 3 failed |
| F | inert declaration + blank glyph | 1 failed (was **6 passed** before the fix) |
| G | colour hidden in the `border` shorthand | 3 failed (`..._not_colour_alone` was **green** before the fix) |

All five mutants left the remaining 1228–1231 tests passing, so the new guards are what catch them.

## Verification

`1235 passed`; `ruff check` clean.

## Known limitation

`.badge` (line 88) and `.badge-accepted` (line 100) have equal specificity, so `border-style: dashed` wins on source order alone. Reordering the badge block would silently drop the dashed border and leave only the glyph — and the guard would still be green, because its assertion is a disjunction that the glyph alone satisfies. The practical blast radius is small, but it is a real property of this guard rather than something the tests rule out.

## Follow-up (noted, not filed)

The `_declarations()` helper ignores `@media` nesting, and the same helper is duplicated across `tests/test_term_css.py`, `tests/test_theme.py`, and this file. The light palette already lives inside a media query, so if any of these rules moved into one, all three files would quietly parse `{}` and pass.

### DIFF
--- a/tests/test_badge_actor_css.py
+++ b/tests/test_badge_actor_css.py
@@ -1,0 +1,173 @@
+# SPDX-License-Identifier: MPL-2.0
+"""Regression lock on the accepted-vs-confirmed badge (#221).
+
+`confirmed` and `accepted` are different events: `accept_fact()` records a person's
+call (actor defaults to the user), `auto_accept_fact()` records a rule's
+(`actor="rule"`, with a rule_name). Both land in `ENGINE_STATUSES`, so a
+rule-accepted fact reasons with exactly the authority of a human-accepted one --
+yet app.css styled them with a single shared rule and fact_row.html shows no actor,
+so the review screen gave the reader no way to tell which had happened.
+
+These tests pin the distinction the way #226 demands it be pinned:
+
+1. **The two badges are genuinely different rules**, compared by parsed declaration
+   set rather than "a rule exists" -- two rules with identical bodies would still be
+   the bug.
+2. **The difference is not colour alone.** Colour properties are stripped before the
+   comparison, so "fix" the bug by re-tinting one badge and this stays red. #226 is
+   about verdicts that depend on colour alone; this is the guard that stops #221's
+   fix from reopening it.
+3. **Both stay in the ok tone**, which is what the issue asked for -- the actor is
+   the difference, not the verdict.
+4. **The glyph carries empty alt text**, so it never leaks into a screen reader's
+   reading of the badge (the badge text already says "accepted").
+5. **`.badge-confirmed` carries no glyph.** That class doubles as a generic ok chip
+   ("covered" on the dashboard, "OK" in the report), so an approval mark there would
+   land on text that is not an accept decision at all.
+6. **The template still emits `badge-{{ status }}`**, without which the CSS is dead.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+WEB = Path(__file__).resolve().parents[1] / "verinote" / "web"
+CSS_PATH = WEB / "static" / "app.css"
+TEMPLATES = WEB / "templates"
+
+# Properties that carry colour and nothing else. Stripped before the "are they really
+# different" comparison so a colour-only distinction cannot satisfy it.
+COLOUR_PROPERTIES = ("color", "border-color", "background")
+
+
+def _read_css() -> str:
+    return CSS_PATH.read_text(encoding="utf-8")
+
+
+def _strip_comments(css: str) -> str:
+    return re.sub(r"/\*.*?\*/", "", css, flags=re.DOTALL)
+
+
+def _declarations(selector: str, css: str) -> dict[str, str]:
+    """Parsed `property -> value` for the flat rule whose selector is exactly `selector`.
+
+    Returns `{}` when no such rule exists -- so the pre-fix state, where neither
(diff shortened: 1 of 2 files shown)


justinjoy (approved):
검토 결과 blocking issue는 없습니다.

`confirmed`와 `accepted`의 상태 의미 차이를 전역 badge CSS에서 non-colour 채널(dashed border, decorative gear glyph)로 구분하면서 두 상태 모두 ok tone을 유지하는 것을 확인했습니다. `.badge-confirmed`에 glyph를 넣지 않은 것도 dashboard/report의 generic ok chip 재사용 때문에 타당합니다. 새 CSS regression tests와 GitHub CI가 모두 통과했습니다.

PR #316, opened 2026-07-20
fix(#220): tell the Ask VERIFIED and UNVERIFIED verdicts apart

### LINKED ISSUE #220: UI: Ask renders a VERIFIED and an UNVERIFIED answer identically
## 문제

Ask 탭의 세 라우트가 **화면에서 구분되지 않는다.**

- `verinote/web/templates/ask.html:18` 이 `<section class="ask-result ask-{{ result.route }}">` 로 훅 클래스를 뱉지만, `app.css` 에 `.ask-engine` / `.ask-fallback` **규칙이 아예 없다** (죽은 훅).
- 라우트 라벨은 양쪽 다 맨 `<h2>` (`ask.html:19`).
- 답변 블록은 `.answer-box pre` (`app.css:109-110`) — 라우트와 무관하게 같은 회색 `--panel` 배경.

즉 엔진이 증명한 답과, 엔진이 답하지 못해 LLM 이 발췌로 지어낸 답이 **같은 모양으로 나온다.** 이 저장소가 최악으로 규정한 실패("조용히 틀린 ok")가 프레젠테이션 계층에서 그대로 일어난다. PR #192 가 evidence-first 출력 계약을 문서화했지만, UI 는 그 계약의 **순서만** 지키고 **판정 구분**은 지키지 않는다.

## 함정: `route` 로는 구분할 수 없다

`VERIFIED — engine` 과 `VERIFIED — engine (negative)` 은 **둘 다 `route == "engine"`** 이다 (`pipeline/ask.py:99` vs `:116`, `:131`). `.ask-{{ route }}` 클래스로는 2종밖에 못 만든다. 분기는 `result.status`(`no_answer`) 또는 `result.label` 로 해야 한다.

## 고칠 방향

답변 위에 판정 배너를 두고, 답변 블록 테두리도 톤을 따르게 한다:

| 라벨 | 톤 | 글리프 |
|---|---|---|
| `VERIFIED — engine` | ok | ✓ |
| `VERIFIED — engine (negative)` | ok | ∅ (검증된 부재) |
| `UNVERIFIED — source exploration` | **warn** | ⚠ |

UNVERIFIED 를 danger 가 아니라 warn 으로 두는 이유: 틀렸다는 판정이 아니라 **판정의 부재**이므로 "불충분"이 정확한 어휘다. 초록·체크·VERIFIED 와 동형인 배너는 금지.

## 데이터 제약 (템플릿이 볼 수 있는 것)

`web/app.py:1038-1043` 이 넘기는 건 `{question, result, error}` 뿐이다. `result.grounding_facts` 에는 **fact id 가 없다** — 근거 사실을 provenance 로 링크하려면 `AskResult` 변경이 필요하다(이 이슈 범위 밖). fallback 라우트는 `query_dl` 이 `None` 이고 `answer`/`evidence` 가 비어 있으므로, 컬럼을 무조건 그리면 빈 칸이 된다.

## 검증

엔진이 답하는 질문과 답하지 못하는 질문을 각각 던져 두 화면이 **색·글리프·테두리에서 다르게** 보일 것. 색맹 시뮬레이션에서도 구분될 것(색 단독 의존 금지).

SeoyunL:
Closes #220

Ask rendered a VERIFIED answer, a VERIFIED negative and an UNVERIFIED fallback identically.

## The discriminator is `status`, not `route`

The issue describes "three routes", but `route` only ever takes two values. All five construction sites in `pipeline/ask.py`:

| line | route | label | status |
|---|---|---|---|
| 76 | `fallback` | UNVERIFIED — source exploration | `empty` |
| 109 | `engine` | VERIFIED — engine | `translated` |
| 125 | `engine` | VERIFIED — engine (negative) | `no_answer` |
| 139 | `engine` | VERIFIED — engine (negative) | `no_answer` |
| 269 | `fallback` | UNVERIFIED — source exploration | `fallback` |

So `route` cannot separate the two VERIFIED verdicts — they are both `engine`. The template now folds `(route, status)` into three verdict tones. All five sites are covered, and there is no silent VERIFIED fallthrough.

The old hook was `ask-{{ result.route }}`, and neither `.ask-engine` nor `.ask-fallback` had a rule anywhere in the stylesheet, which is why all three answers looked the same.

## Three channels

| verdict | glyph | border style | colour |
|---|---|---|---|
| verified | `✓` | solid | `--ok` |
| verified-negative | `∅` | double | `--ok` |
| unverified | `⚠` | dashed | `--warn` |

Colour alone would repeat #226, so the glyph and the border style carry the verdict independently — the border style is what survives greyscale and colour-vision deficiency. Glyphs use the `content: "✓" / ""` empty-alt convention from `.term-term`, because the label text already states the verdict and announcing "check mark" ahead of "VERIFIED" is noise. The `<h2>` is kept and only gains a class: it has sibling `<h2>`s in the same section, so demoting it would break the outline, and it is the only path by which the verdict reaches a screen reader.

The answer body is toned too, but it answers a narrower question — "can I trust this" (ok vs warn). Telling positive from negative is the banner's job. Splitting the two engine verdicts there would mean a `double` at 3px, which renders 1/1/1 and reads as a plain solid: an invisible distinction the tests would then lock in.

## Why nine commits

The production fix is two commits and has been unchanged since `27af8e1`. The other seven are the regression lock, which took five rounds because the same defect kept reappearing one layer deeper: **the guards asserted the traces of the prescription rather than what it renders.**

1. **`721073f`** — the fix: key off `(route, status)`.
2. **`4d7ba57`** — 3px `double` renders as `solid`, but the guard counted it as different. It was locking a distinction nobody could see.
3. **`b472f84`** — the answer body only asserted "a tone rule exists". Collapsing the three rules into one made an LLM-written fallback answer wear the engine's tone, and nothing caught it. That had been red before only by accident, and a legitimate refactor removed the accident.
4. **`27af8e1`** — the shape channel could vanish unguarded: unifying only the banner's `border-left-style` was green.
5. **`567f6d9` / `3e581ad` / `0a65fb9`** — all three channel guards compared declaration *text*. `font-style: normal` (a no-op on a `<pre>`) faked a difference; `solid`/`hidden`/`none` counted as three border styles though two paint nothing; an empty glyph passed; a zeroed skeleton width erased every border while the styles still read as distinct.
6. **`7a862a5`** — the turn. Rounds 1–5 all answered by **widening the set of values recognised as inert**, which narrowed the preimage without closing it: every property has a value that happens to match its context (`margin-left: 0` against a base `margin: .6rem 0 1rem`). That enumeration has no end. So the trust boundary **stopped judging what is inert and started reading which signals are painted** — the comparison ignores anything bolted on beside them, which makes the enumeration unnecessary rather than longer.
7. **`0b0d922`** — scoping the docstring to what the guard actually removes.

## Evidence

Mutations were run against `git archive` copies, never the working tree.

**Immunity.** With the unverified answer body made pixel-identical to the verified one, twelve arbitrary added declarations are each red: `margin-left: 0`, `letter-spacing: 0`, `font-style: normal`, `line-height: 1.5`, `padding-left: 0`, `word-spacing: 0`, `text-indent: 0`, `opacity: 1`, `transform: none`, `z-index: auto`, `tab-size: 4`, `cursor: text`.

**Not over-restrictive.** Eight legitimate non-colour implementations stay green: an underline alone, a `box-shadow` alone, italic alone, a background image alone, a different glyph with `dotted`, `rem` units, an added `font-weight`, and a differing width.

**Per-channel isolation.** Killing one channel goes red on its own: glyph → the glyph guard, border style → the shape guard, trust boundary → the boundary guard. The reverse direction is covered too — forcing the two engine verdicts apart is red.

`1240 passed`, `ruff check` clean.

## What these guards do not catch

Worth stating plainly, since the point of this PR is guards that mean what they say.

- A **signal property carrying a value that is inert in this context** still passes: `font-weight: 400` on a `<pre>` that already inherits 400, a zero-geometry `box-shadow`, an `outline-width` with no `outline-style`, `text-decoration-line: none` after an underline. This is the residual preimage. It is finite — nine signal properties — rather than "any CSS property", and reaching it takes **two** steps, because the verdict's own channel has to be collapsed first and that alone is red.
- **Values are not normalised**: `3.0px` and `3px` compute identically but compare as two.
- The guards read declarations, not a rendered box, so a property neutralised by its context (`border-right-style` with no right-hand width) still counts. Closing that needs a real layout engine.
- A distinction carried **only** by `filter: grayscale(1)` fails **red**, because `filter` is not in the signal set. That is the safe direction: the guard says so rather than passing silently.

## Follow-up (not filed)

`UNVERIFIED — engine unavailable` — separating "the engine legitimately has no answer" from "the engine is broken" needs a failure-reason field on `AskResult`; the template cannot sniff `result.reason` for it. The root cause is #237. The tone keys are `ask-verdict-*`, so it drops in as one CSS rule and one template branch.

### DIFF
--- a/tests/test_ask_verdict.py
+++ b/tests/test_ask_verdict.py
@@ -65,7 +65,13 @@ RULE = re.compile(r"(?P<selector>[^{}]+)\{(?P<body>[^{}]*)\}")
 # A declaration is "colour" if its property names a colour or its value reaches for a
 # palette token. Stripping these is the greyscale simulation guard 3 runs.
 COLOUR_PROPERTY = re.compile(r"^(color|background|border(-[a-z]+)*-color|outline-color)$")
-COLOUR_TOKEN = re.compile(r"--(ok|warn|danger|accent|line|muted|panel|fg|bg|term)")
+# A palette reference *inside* a value. Only the reference is removed, not the whole
+# declaration: a `box-shadow`'s offsets are a shape channel even though its colour is not,
+# and dropping the declaration wholesale would reject a shadow-only design as
+# "colour alone" when its geometry is what carries the distinction.
+COLOUR_VAR = re.compile(
+    r"var\(\s*--(?:ok|warn|danger|accent|line|muted|panel|fg|bg|term)[a-z-]*\s*\)"
+)
 
 # Below here the tests stop trusting that a declaration *exists* and start asking whether
 # it draws. Counting declarations lets a rule fake a difference with something inert:
@@ -158,11 +164,20 @@ def _tone_declarations(tone: str) -> set[tuple[str, str]]:
 
 
 def _without_colour(declarations: set[tuple[str, str]]) -> set[tuple[str, str]]:
-    return {
-        (prop, value)
-        for prop, value in declarations
-        if not COLOUR_PROPERTY.match(prop.split("|")[-1]) and not COLOUR_TOKEN.search(value)
-    }
+    """The greyscale simulation: colour properties dropped, colour references erased.
+
+    Erasing the reference rather than the declaration keeps whatever shape the value
+    still carries. `box-shadow: inset 3px 0 0 0 var(--warn)` becomes `inset 3px 0 0 0`,
+    which is a real channel; only a value that was *nothing but* colour disappears.
+    """
+    kept = set()
+    for prop, value in declarations:
+        if COLOUR_PROPERTY.match(prop.split("|")[-1]):
+            continue
+        stripped = " ".join(COLOUR_VAR.sub(" ", value).split())
+        if stripped:
+            kept.add((prop, stripped))
+    return kept
 
 
 def _drawn(prop: str, value: str) -> str | None:
--- a/tests/test_ask_verdict.py
+++ b/tests/test_ask_verdict.py
@@ -1,0 +1,248 @@
+# SPDX-License-Identifier: MPL-2.0
+"""Regression lock on the Ask verdict rendering (#220).
+
+The Ask page hung its result styling off `ask-{{ result.route }}`, but `route`
+only ever takes two values (`engine` and `fallback`) while the page has three
+verdicts to tell apart: a verified answer, a verified *negative* ("no confirmed
+facts match"), and an unverified fallback. `route` collapses the first two --
+and `.ask-engine` / `.ask-fallback` had no rule anywhere in app.css, so all
+three rendered identically. The discriminator for the two engine verdicts is
+`status`.
+
+Locking "the template emits three classes" would be too weak, so these tests pin
+the whole path, and pin *how* the three are told apart:
+
+1. The three verdicts render three distinct hook classes.
+2. Every hook the template can emit is styled, and app.css carries no hook the
+   template never emits -- the dead-hook bug this issue is about.
+3. The tones still differ once the colour is stripped out of their declarations.
+   Colour alone is what reopened #226; a border style survives greyscale and
+   colour-vision deficiency.
+4. The three glyphs differ, and each is announced as nothing.
+5. Every label `pipeline/ask.py` can produce is covered, so a verdict added
+   upstream cannot silently inherit another verdict's rendering.
+
+CSS is compared as *parsed declaration sets*, not as "does a rule exist": two
+rules with identical bodies are still the bug. Rendering goes through a jinja2
+overlay rather than the app, so a stub `base.html` keeps this off the real one.
+"""
+
+from __future__ import annotations
+
+import re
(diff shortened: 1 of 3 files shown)


justinjoy (approved):
검토 결과 차단 이슈는 없습니다. Ask 결과의 route/status 조합이 세 verdict 클래스로 분리되고, 색상 외 glyph/border-style 채널 및 verified/unverified answer body 구분에 대한 회귀 테스트가 포함되어 있습니다. 로컬에서 tests/test_ask_verdict.py, tests/test_ask.py, tests/test_web.py를 확인했습니다.

PR #317, opened 2026-07-20
fix(#261): order answer buckets by question number, not by digit sort

### LINKED ISSUE #261: fix: 답변 qid 가 사전순 정렬이라 질문이 10개를 넘으면 q10 이 q2 앞에 온다
**위치**: `engine/duckdb_backend.py:396-399`, `engine/wirelog.py:317-318` (양 백엔드 동일 패턴)

```python
answers = [
    f"q{qid}: {', '.join(sorted(set(vals)))}"
    for qid, vals in sorted(answers_by_q.items())   # <- 문자열 정렬: "10" < "2"
]
```

## 현상

qid 는 `answer_q<id>` 접두사에서 잘라낸 문자열이라 `sorted()` 가 사전순으로 동작. 질문 id 가 10 이상이 되는 순간 리포트의 answers 섹션 순서가 `q1, q10, q11, q2, ...` 로 뒤섞인다. 양 백엔드가 같은 버그를 공유하므로 백엔드 간 불일치는 아니고 표시 순서 버그.

## 제안

`sorted(answers_by_q.items(), key=lambda kv: int(kv[0]))`. qid 는 파서가 `answer_q[0-9]+` 로 강제하므로(`wirelog.py:199` `_ANSWER_DECL`) `int()` 는 안전.

## 충돌 주의

열린 PR #253(#167, tuple 기반 dedup)이 정확히 이 라인들을 만진다. #253 머지 후 그 위에서 수정하거나 #253 에 얹는 게 낫다.

## 테스트

질문 12개(q1~q12)를 가진 리포트에서 answers 순서가 수치순인지 단언.

SeoyunL:
Closes #261

The answers section sorted qid **strings**, so past nine questions `q10` came before `q2`. Both engine backends shared the bug.

## The issue's safety claim is wrong — this is the core of the PR

The issue says the parser enforces `answer_q[0-9]+`, so `int()` is safe. It is not. `_ANSWER_DECL` only checks the **LLM-proposed query snippet** inside `validate_query()`; both answer collectors filter the combined `policy_dl + query_dl` program with nothing but `startswith("answer_q")`. **A policy is user-authored**, so a non-numeric qid does reach the sort. Measured: a policy declaring `.decl answer_qfoo(value: symbol)` returns `ok=True, answers=['qfoo: London']` today.

Taking the issue's proposed `key=lambda kv: int(kv[0])` would have broken policies that pass right now:

- **duckdb**: the collector runs inside `except Exception -> _internal_engine_error`, so the report becomes `ok=False` with `"ERROR internal engine error: invalid literal for int()"` — **the review gate flips**.
- **wirelog**: the comprehension sits outside any `try` (the `try` wraps only the pyrewire session), so the `ValueError` traceback escapes.

Fixing a display-order bug would have taken down the verdict itself. The reviewer confirmed both failure modes by running them independently.

## The fix

One shared key function in `wirelog.py`, used at both sort sites:

```python
_NUMERIC_QID = re.compile(r"[0-9]+\Z")

def answer_bucket_sort_key(qid: str) -> tuple[int, int, str]:
    if _NUMERIC_QID.fullmatch(qid):
        return (0, int(qid), qid)
    return (1, 0, qid)
```

- Sort sites: `engine/duckdb_backend.py` and `engine/wirelog.py`. duckdb already imports from wirelog, so there is **no new module and no change to the dependency direction**; the key function has exactly one owner.
- **Render format `q{qid}:` is unchanged** — only the sort key moved. Zero migration. Zero-padding was rejected: qid is `questions.id`, an INTEGER primary key, and `ask.py`'s `^q\d+:\s*` plus ~20 existing assertions are tied to the current format.
- **`str.isdigit()` is deliberately not used**: `"²".isdigit()` is True but `int("²")` raises. `[0-9]+` matches the "numeric qid" definition `_ANSWER_DECL` and `_ANSWER_RE` already use, so the two cannot drift.
- **Why `q12abc` sorts as non-numeric rather than parsing a `12` prefix**: prefix-parsing would fold `answer_q12` and `answer_q12abc` into **the same bucket, merging answers from two different predicates under one `q12:` label** — worse than a wrong order. qid renders a question's INTEGER PK, and `12abc` is not any PK. Placing it after the numbered buckets is deterministic and safe.

## Regression guards — `tests/test_answer_qid_order.py` (new file)

Every ordering guard uses **more than nine questions** (1..12, and the boundary 9/10/99/100/101). With q1..q9 string order and numeric order agree, so a smaller fixture would pass against the very bug being fixed. Guards assert the **actual output order**, not that a sort function was called.

### Mutation evidence

| # | Mutant | Result |
|---|---|---|
| M1 | revert the duckdb sort site **only** | 3 failed, 5 passed — duckdb guards red, **wirelog guards stay green** |
| M2 | revert the wirelog sort site **only** | 3 failed, 5 passed — wirelog guards red, **duckdb guards stay green** |
| M3 | delete the non-numeric branch | 4 failed (`invalid literal for int(): 'foo'`) |
| M4 | **the issue's proposed naive `int()`** at both sites | 4 failed, 4 passed — all four ordering guards green, **only the non-numeric guards red** |
| M5 | swap the numeric / non-numeric ranks | 2 failed (`At index 0 diff: 'q12abc' != 'q1'`) |

**M1 and M2 staying green for each other** is the proof that the two sort sites are locked independently — a guard set that passes when only one site is fixed would not protect both.

**M4 is the sole proof that the non-numeric guards earn their place**: the ordering guards cannot tell the difference, only the non-numeric ones can.

## The second lesson: the anchor guard

Review found that loosening `_NUMERIC_QID` to `r"[0-9]+"` with `.match` revives `ValueError: invalid literal for int() with base 10: '12abc'` — **while all eight guards stayed green.** The non-numeric guards were all purely alphabetic (`foo`/`bar`), and an alphabetic qid never reaches the branch the anchor defends. Five mutations had passed because every one of them touched a line of code.

> **Guard strength is input coverage x assertion strength, and line-level mutation audits only the second factor.** Anchors are especially exposed: without one, nothing about the output changes until that single shape arrives, so even outcome-based guards miss it.

The second commit adds `q12abc` alongside the existing `foo`/`bar` (it does not replace them — M3 still fails on `'foo'`, so the all-alphabetic case remains covered). With it, the loosened-anchor mutant goes red on four guards.

## Verification

- `pytest tests -q -rs` -> **1237 passed**, no skip section (skip 0)
- `ruff check .` -> All checks passed

## Follow-up (not filed here)

The qid format for `answer_q` is defined **twice and inconsistently**: the engine collectors accept anything via `startswith("answer_q")`, while the contract (`_ANSWER_DECL`, `_ANSWER_RE`) requires `answer_q[0-9]+\Z`. The non-numeric fallback branch in this PR is a downstream symptom of that split, as is the fact that a non-numeric qid is silently dropped from Traceability while a `qfoo: ` prefix leaks to the user. Whether a non-numeric `answer_q*` is an answer at all or a policy-validation error is a contract question, not a display-order one, and is out of scope here. It is being filed separately as a single issue so the split itself gets addressed rather than one symptom at a time.

### DIFF
--- a/tests/test_answer_qid_order.py
+++ b/tests/test_answer_qid_order.py
@@ -72,10 +72,13 @@ def test_wirelog_answers_order_across_the_hundreds_boundary():
 # from reaching the answer collectors. Ordering it must not cost the report its
 # `ok` (duckdb turns the exception into an internal engine error, which flips the
 # review gate) nor raise out of wirelog, which collects outside any try.
-_NON_NUMERIC_POLICY = (
-    _RELATION_DECL + ".decl answer_qfoo(value: symbol)\n"
-    'answer_qfoo(O) :- relation("Ada", "born_in", O).\n'
-)
+#
+# `12abc` is here for the anchor: a qid that *starts* numeric is the only shape an
+# unanchored `[0-9]+` match still calls numeric, and `int("12abc")` then raises.
+# All-alphabetic qids like `foo` never reach that branch, so they cannot cover it.
+_NON_NUMERIC_QIDS = ["foo", "12abc"]
+_NON_NUMERIC_POLICY = _RELATION_DECL + _query_for(_NON_NUMERIC_QIDS)
+_NON_NUMERIC_ANSWERS = ["q12abc: London", "qfoo: London"]
 
 
 def test_duckdb_keeps_a_non_numeric_answer_predicate_clean():
@@ -83,7 +86,7 @@ def test_duckdb_keeps_a_non_numeric_answer_predicate_clean():
     rep = run_check_duckdb(_FACTS, policy_dl=_NON_NUMERIC_POLICY)
 
     assert rep.ok is True
-    assert rep.answers == ["qfoo: London"]
+    assert rep.answers == _NON_NUMERIC_ANSWERS
 
 
 def test_wirelog_keeps_a_non_numeric_answer_predicate_clean():
@@ -91,18 +94,13 @@ def test_wirelog_keeps_a_non_numeric_answer_predicate_clean():
     rep = _wirelog_check(policy_dl=_NON_NUMERIC_POLICY)
 
     assert rep.ok is True
-    assert rep.answers == ["qfoo: London"]
+    assert rep.answers == _NON_NUMERIC_ANSWERS
 
 
-_MIXED_POLICY = (
-    _RELATION_DECL
-    + _query_for(range(1, 13))
-    + ".decl answer_qfoo(value: symbol)\n"
-    'answer_qfoo(O) :- relation("Ada", "born_in", O).\n'
-    ".decl answer_qbar(value: symbol)\n"
-    'answer_qbar(O) :- relation("Ada", "born_in", O).\n'
+_MIXED_POLICY = _RELATION_DECL + _query_for(
+    list(range(1, 13)) + ["foo", "bar", "12abc"]
 )
-_MIXED_ORDER = [f"q{n}" for n in range(1, 13)] + ["qbar", "qfoo"]
+_MIXED_ORDER = [f"q{n}" for n in range(1, 13)] + ["q12abc", "qbar", "qfoo"]
 
 
 def test_duckdb_sorts_numeric_answers_before_non_numeric_ones():
--- a/tests/test_answer_qid_order.py
+++ b/tests/test_answer_qid_order.py
@@ -1,0 +1,119 @@
+# SPDX-License-Identifier: MPL-2.0
+"""Answer buckets are ordered by question number, not by how the digits sort.
+
+Every ordering guard here uses more than nine questions on purpose: with q1..q9
+string order and numeric order agree, so a smaller fixture would pass against
+the very bug this file locks.
+"""
+
+import pytest
+
+from verinote.engine.duckdb_backend import run_check_duckdb
+from verinote.engine.wirelog import compile_dl, run_check
+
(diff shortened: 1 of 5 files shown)


justinjoy (changes requested):
차단 이슈가 있습니다.

`answer_bucket_sort_key()`가 `[0-9]+` qid를 모두 `int(qid)`로 변환하고 있어, 매우 긴 숫자 suffix를 가진 사용자 작성 `answer_q...` predicate에서 Python의 정수 문자열 변환 한도(`sys.get_int_max_str_digits()`, 기본 4300자리)에 걸려 `ValueError`가 발생합니다. 이 predicate는 `answer_q` 수집 경로까지 도달할 수 있으므로, 이번 PR이 `12abc`에 대해 막으려는 malformed/user-authored qid로 인한 engine failure가 긴 숫자 qid에서는 여전히 남습니다.

재현 예: `qid = "1" * 5000`으로 `.decl answer_q<qid>(value: symbol)`와 대응 rule을 policy에 넣으면 `wirelog.run_check(...)`가 answer 정렬 중 예외를 전파합니다. DuckDB backend도 같은 `answer_bucket_sort_key()`를 사용하므로 동일한 정렬 경로의 영향을 받습니다.

긴 숫자 suffix 회귀 테스트를 추가하고, 무제한 `int()` 변환 없이 숫자 문자열을 정렬하거나 변환 실패 시 malformed qid와 같은 뒤쪽 bucket으로 보내도록 수정해 주세요.

PR #302, opened 2026-07-19
docs(#279): document how to update and re-pin the vendored htmx

SeoyunL:
Closes #279

## What

Documents how to update the vendored htmx and re-pin its integrity hash, in a new
`docs/vendored-assets.md`, plus a one-line link from README's Contributing section.

**Documentation only — no changes under `verinote/**` or `tests/**`.** Actually
bumping htmx is deliberately out of scope (see Follow-ups).

## Why the document centres on two lines

The integrity check is not an SRI attribute, a checksum file, or a manifest. It is
a single test constant, `HTMX_SHA256` in `tests/test_base_template_assets.py`, and
the only record of *which* version is vendored is the `# Source:` comment directly
above it. So an update has to change **both of those lines together**, and that is
what the procedure is built around.

**The trap in the same file:** its module docstring also mentions `2.0.3`, but that
is the #219 regression narrative recording what base.html loaded *before* vendoring.
It must not be updated — rewriting it would assert that base.html once loaded a
version it never loaded, destroying the explanation of why the guards exist. The
document quotes it verbatim and says to leave it alone.

**The `/releases/latest` trap:** htmx marks its prereleases with `prerelease: false`,
so the API returns `v4.0.0-beta5` as "latest" (measured against the API on
2026-07-20). Version selection therefore comes *before* the fetch step, so a
maintainer reading top-to-bottom meets the warning before overwriting anything.

The procedure covers: choosing the version, fetching the release artefact,
cross-checking the bytes against an independent mirror, updating the pin, verifying,
manual UI check, and committing bytes and pin as one commit. Rollback instructions
are included at both points where the procedure can abort with an unverified blob
already on disk. Cadence and the signals to watch (releases, GitHub Advisory,
`htmx.org` npm advisory, plus a quarterly manual pass) are covered.

## Verification

Every command in the document was executed, not just written:

- GitHub release asset, unpkg, and jsDelivr are byte-identical for 2.0.3, all
  reproducing the pinned `491955cd…784e`.
- The procedure generalises to v2.0.9 (the newest stable 2.x).
- The rollback command actually restores: the vendored file was dirtied, then
  `git checkout --` returned it to the pinned digest and a clean status.
- `1229 passed`, `ruff check .` clean.

## Review history

Eight findings from two review rounds plus an architect critique are folded in.
The most significant: the first draft asserted that grepping the repository for a
version number "will find nothing else", which was **false** — there is a second
`2.0.3` in the same test file's docstring. Left in, it would have led a reader to
[...]

PR #318, opened 2026-07-20
build(#306): update vendored htmx from 2.0.3 to 2.0.9

### LINKED ISSUE #306: build: vendored htmx 를 2.0.3 → 2.0.9 로 갱신 (#279 절차 첫 적용)
vendored htmx 가 **2.0.3** 인데 안정 2.x 는 이미 **2.0.9** 다 (2.0.4~2.0.9, 6 릴리스 뒤처짐). PR #302 가 문서화한 갱신 절차를 실제로 한 번 돌리는 작업이다.

## 절차

`docs/vendored-assets.md` 의 5단계를 그대로 따르면 된다. 요지:

1. **버전 선정** — `/releases/latest` 를 믿지 마라. htmx 는 프리릴리스에도 `prerelease: false` 를 달아서 오늘 기준 `v4.0.0-beta5` 를 반환한다. 태그 목록에서 2.x 안정 최신을 직접 골라야 한다.
2. 취득 → 독립 대조(GitHub release asset / unpkg 바이트 동일 확인) → `HTMX_SHA256` 과 바로 위 `# Source:` 주석을 **함께** 갱신 → `pytest` → 수동 UI 확인.
3. **바이트와 해시 갱신은 한 커밋에.** 쪼개면 중간 리비전이 red 라 `git bisect` 가 false failure 를 얻는다.

## 주의

`tests/test_base_template_assets.py` 상단 docstring 의 `2.0.3` 은 vendoring **이전** 상태를 기록한 #219 회귀 서사다. **갱신하지 마라.** 갱신하면 "base.html 이 한 번도 로드한 적 없는 버전을 로드했었다"는 문장이 되어 그 테스트의 존재 이유가 무너진다.

## 범위

`verinote/web/static/htmx.min.js` + `tests/test_base_template_assets.py`. 메이저 라인을 넘지 않으므로 템플릿 변경은 없어야 한다.

Refs #279, #219

SeoyunL:
Updates the vendored htmx from 2.0.3 to 2.0.9, following the 5-step procedure in
`docs/vendored-assets.md`. This is the procedure's first real application.

> **Merge order: #302 first.** That procedure document is still unmerged — PR #302 is
> open, so `docs/vendored-assets.md` exists neither on `origin/main` nor on this branch.
> The two PRs touch disjoint files and do not conflict, but "first application of the
> #279 procedure" only reads as true once #302 has landed.

Two files change. The vendored bytes and the pin that guards them move together in one
commit — splitting them would leave an intermediate revision where the hash test is red,
which would hand `git bisect` a false failure. A second, comment-only commit records why
the pin is 2.0.9.

- `verinote/web/static/htmx.min.js` — 50387 -> 51332 bytes
- `tests/test_base_template_assets.py` — the `# Source:` comment, `HTMX_SHA256`, and a
  comment recording the version choice

The `2.0.3` in that file's module docstring is left alone on purpose: it records the
pre-vendoring state that #219 fixed. Updating it would turn the docstring into a claim
that base.html once loaded a version it never loaded.

No template and no CSS changes. In particular `app.css` is untouched — see below.

## Why 2.0.9 and not 2.0.10

2.0.10 exists and is npm's `dist-tags.latest`. It was considered and deliberately not
taken, so nobody has to repeat the investigation. This is also recorded next to the pin
in `tests/test_base_template_assets.py`, so the reasoning travels with the code.

- **Decisive: 2.0.9 was not a preference, it was the only version the procedure admits.**
  Step 2 of #306 requires confirming identical bytes from the GitHub release asset *and*
  unpkg. Neither `v2.0.10` nor `v2.0.8` has a GitHub Release — `releases/tags/v2.0.10`
  returns 404 — so the mandated two-source comparison cannot be performed against them
  at all.
- 2.0.10 carries exactly two changes, both inert here. It restores TypeScript type
  definitions, and we vendor only the `min.js` bytes. And it wraps a settle lookup in
  `CSS.escape()`; `schema.sql:99` declares `id INTEGER PRIMARY KEY`, so a selector like
  `#fact-123` is always CSS-safe.
- It is not a security patch.

## Supply-chain verification

The 2.0.9 bytes were fetched from three independent publishers and are byte-identical.
The comparison was then re-run independently during review and reproduced the same digest:

| Publisher | SHA256 | Bytes |
|---|---|---|
| GitHub release asset | `57d9191515339922bd1356d7b2d80b1ee3b29f1b3a2c65a078bb8b2e8fd9ae5f` | 51332 |
| unpkg | `57d9191515339922bd1356d7b2d80b1ee3b29f1b3a2c65a078bb8b2e8fd9ae5f` | 51332 |
| jsDelivr | `57d9191515339922bd1356d7b2d80b1ee3b29f1b3a2c65a078bb8b2e8fd9ae5f` | 51332 |

The installed file was re-hashed after the copy and matches. `version:"2.0.9"` is present
in the payload, and the `/static` mount was confirmed to serve bytes identical to the file
on disk (the existing test only asserts 200 plus a JS content-type).

## Upstream changes that touch verinote

verinote uses eight `hx-*` attributes: `hx-get`, `hx-post`, `hx-target`, `hx-swap`,
`hx-select`, `hx-trigger`, `hx-indicator`, `hx-disabled-elt`. A repo-wide search finds no
use of `hx-boost`, `hx-push-url`, `hx-history`, `HX-Location`, `hx-on`, `hx-sync`,
`hx-swap-oob`, `hx-vals`, `hx-confirm`, `htmx.ajax`, or `HX-Redirect`, so the 2.0.4
`htmx.ajax` retarget, the 2.0.5 history/sessionStorage move, the 2.0.8 `hx-sync` and
Shadow DOM work, and the 2.0.9 `HX-Location` replace all land on unused surface.

Response headers are a second contact axis, and verinote does use one: `app.py:383` emits
`HX-Refresh`. It is unchanged in 2.0.9 and still maps to `location.reload()` — see the
follow-up note at the end.

Two changes are active, and both are improvements for this codebase:

- **2.0.7** fixed `htmx-swapping` lingering after an `outerHTML` swap. `outerHTML` is
  verinote's primary swap mode.
- **2.0.9** made `hx-disabled-elt` preserve elements that were already disabled.
  `settings.html:54` uses `hx-disabled-elt`.

2.0.8 routes `parseHTML` through `Document.parseHTMLUnsafe()`. The 2.0.9 bytes contain a
`"parseHTMLUnsafe" in Document` feature guard with a `DOMParser` fallback, so there is no
regression on older browsers.

**`app.css` is intentionally not modified.** 2.0.7 added `visibility` to htmx's injected
indicator styles. Read from the bytes: 2.0.9 injects `opacity:0;visibility:hidden` (2.0.3
injected `opacity:0` alone), while `app.css:193-195` drives the same elements with
`display:none` / `display:inline`. Three distinct properties, so the rules coexist rather
than conflict.

## Automated verification

- Full suite: **1229 passed**. `ruff check .` clean.
- The hash pin is non-vacuous, and each half fails independently:
  reverting only the bytes -> `test_vendored_htmx_bytes_match_the_pinned_hash` fails;
  reverting only the pin -> the same test fails; restored -> 5 passed.
- No template in the repo references an external asset origin.

## Manual verification

The hash test proves only that the file is a specific sequence of bytes; it cannot show
that htmx initialises and that `hx-*` fires. So this was driven through a real browser
(Chrome automation) against this branch's bytes — `verinote ui` serving htmx 2.0.9, serving
hash re-confirmed as `57d9191...784e` / 51332 bytes — on a seeded KB with a 4-item review
queue.

Confirmed:

- **Offline.** Across review, sources, settings and questions, all **13 requests went to
  `127.0.0.1`** — zero external origins. `htmx.min.js` served 200 from `/static/` each time.
- **htmx loads and initialises.** `window.htmx.version === "2.0.9"`, with
  `htmx-internal-data` attached across 12 processed `hx-*` elements.
- **Swap paths:**

| Path | Result |
|---|---|
| `fact_row` toggle | `POST /facts/4/toggle` 200 -> row swapped `candidate` -> `confirmed` |
| `fact_row` accept | `POST /facts/4/accept` 200 -> row swapped |
| `fact_row` edit | `GET /facts/3/edit` 200 -> `editing` class + edit form appears |
| `fact_edit` cancel | `GET /facts/3/row` -> original row restored, edit button back |
| `settings` connection test | `POST /settings/test` -> response handled (502 from an unconfigured provider — an environment condition, not an htmx fault) |

- **The 2.0.7 fix holds.** After every swap, **zero lingering `.htmx-swapping` /
  `.htmx-request`** — checked on toggle, accept, edit and cancel individually. This is the
  point where the version difference would actually show.

Not covered, stated plainly so the gaps are known at merge time rather than assumed away.
Each was closed by a *different* argument — only one of the four is genuinely "same
machine as a path we did exercise", and collapsing them into one blanket claim would
overstate the evidence:

- **`sources` 2-second polling** — not exercised: the seeded KB has no running extraction
  job, so `hx-trigger="every 2s"` never renders. **Zero runtime evidence.** The residual
  risk is bounded only by the changelog: nothing in 2.0.4–2.0.9 touches polling or
  `hx-select`.
- **`hx-disabled-elt` disable/re-enable transition** — **unreachable, not merely
  unobserved.** The form holds exactly one button, so under the pre-disabled condition
  there is nothing to click and nothing to submit, and with no text input there is no
  implicit-submit-on-Enter path either. The 502 also returned too fast to sample the
  transition, but the reachability argument is what closes this.
- **`fact_edit` amend submission** — the one true isomorph: same `hx-post` + `outerHTML`
  swap machinery as toggle/accept, which were exercised. The only version-specific risk
  was 2.0.7's `reportValidity()` change, and that is closed by `reportValidityOfForms`
  defaulting to false plus zero `htmx-config` meta tags in the repo.
- **`questions` body swap** — a *different* mechanism, not an isomorph: `hx-target="body"`
  routes through the 2.0.8 `parseHTML` change. Closed by reading the shipped bytes, which
  carry the `"parseHTMLUnsafe" in Document` feature guard and a `DOMParser` fallback.

## Follow-up (noted, not filed)

Review found that `app.py:383` emits `HX-Refresh` (asserted in four places in
`test_review_gate.py`). The upstream analysis had closed the response-header axis by
looking only at `HX-Location` — a gap worth naming, since 2.0.9 is precisely a release
that touches `HX-Location`. Checked and harmless: the `HX-Refresh` branch still maps to
`location.reload()` and is unchanged in 2.0.9. The durable fix is documentation — the
`HX-*` response headers are a contact surface and should be listed as one in
`docs/vendored-assets.md` (PR #302's file). No issue filed for this.

Closes #306


justinjoy (changes requested):
차단 이슈가 있습니다.

현재 main에는 `docs/vendored-assets.md`가 있고, 이 문서의 “Where the version and the hash live” 섹션은 현재 vendored htmx 버전과 SHA256을 `2.0.3` / `491955...`로 안내합니다. 이 PR은 실제 vendored 파일과 `tests/test_base_template_assets.py`의 pin을 `2.0.9` / `57d919...`로 바꾸지만 문서를 함께 갱신하지 않아, 병합 후 수동 vendoring 절차가 오래된 현재 버전과 해시를 가리키게 됩니다.

PR 브랜치가 문서 추가 전 main에서 갈라져 있어 diff에는 문서 파일이 보이지 않지만, rebase merge 후에는 main의 `docs/vendored-assets.md`가 그대로 남습니다. 따라서 이 PR에서 해당 문서의 현재 버전/hash 예시와 “두 줄이 전체 수정 범위”라는 설명을 새 pin 구조에 맞게 업데이트해 주세요.

참고로 vendored `htmx.min.js` 자체는 unpkg 및 GitHub release asset의 htmx 2.0.9 SHA256(`57d9191515339922bd1356d7b2d80b1ee3b29f1b3a2c65a078bb8b2e8fd9ae5f`)과 일치했고, `tests/test_base_template_assets.py`는 로컬에서 통과했습니다.

justinjoy (approved):
변경요청 사항이 반영되었습니다. vendored htmx 2.0.9 pin/hash와 docs/vendored-assets.md의 현재 버전 예시 및 업데이트 절차가 일치하고, 2.0.3 언급은 #219의 과거 CDN 이슈 설명으로만 남아 있습니다. CI와 로컬 검증 모두 통과했습니다.

Issue #160, opened 2026-07-12
Do not resurrect rejected facts when a source is re-extracted


SeoyunL:
## Symptom

Re-extracting the same source **resurrects facts the human already rejected**. A rejected
(`superseded`) fact comes back as a brand-new `candidate` row in the review queue. Repeat the sync
and duplicates accumulate without bound.

## Repro

```
sync                     -> fact#1 candidate ("Ada born_in Paris")
human rejects            -> fact#1 = superseded
sync the same source again (or re-upload the same file in /sources)
-> review_queue: [fact#2 candidate ("Ada born_in Paris")]     <- the rejection is undone
-> facts: [(1,'superseded'), (2,'candidate')]
```

Web upload is worse: `store_source` overwrites `sources/<name>.txt` and immediately calls
`sync_sources`, so re-uploading the same filename re-inserts everything every time.

Conversely, when the source *text* changes, facts extracted from the **old** text stay `confirmed`
and keep feeding the engine — there is no retire path. Re-upload therefore gives the worst
combination: stale facts kept + duplicate candidates added.

## Cause

`verinote/pipeline/extract.py:34-44` calls `store.add_fact` unconditionally;
`verinote/store/db.py:138-176` always INSERTs (no upsert, no dedup); `schema.sql` has only the
non-unique `idx_facts_triple`, no UNIQUE constraint. `add_source` reuses the source id via
`ON CONFLICT` (`db.py:86-92`), but nothing analogous exists for facts.

## Fix direction

On re-extraction, reconcile against existing rows for the same (source, triple) instead of blindly
inserting: keep the human's decision, and surface a re-extracted-but-previously-rejected fact as an
explicit signal rather than a fresh candidate.

No test covers re-sync/re-upload (`tests/test_pipeline.py`, `tests/test_web.py` extract once).

_Recurrence of factlog's most frequent bug class: regenerating derived state erases the human's decision._

SeoyunL:
**Scope updated after verification against `origin/main` (4cf90da), and absorbing #187.**

Two corrections to the issue as filed:

1. **"Web upload is worse" is no longer true.** The chunked job path already dedupes: `extract.py:237-240` calls `store.fact_exists_for_source(...)` (`db.py:604-617`) and skips existing (source, triple) rows. Reproduced: upload → reject → re-upload the same file → facts stay `[(1,'superseded')]`.

[...]

PR #319, opened 2026-07-20
fix(#240): resume a rolled-back extraction job, and stop reviving superseded ones

### LINKED ISSUE #240: fix: sync 가 롤백된 pending 잡을 재개하지 않고 소스를 통째로 다시 추출한다
## 문제

**halt 롤백 메시지가 시키는 대로 하면 소스가 통째로 재추출된다.**

#194 의 롤백은 사용자에게 이렇게 말한다:

> Rolled back to pending at 1/6 chunk(s). Restore the policy file (or run `verinote policy reset --force`), **then re-run the analysis.**

시키는 대로 `verinote sync` 를 돌리면 — `cmd_sync` 는 그 잡을 **재개하지 않는다.** `cli.py:293` 이 소스마다 무조건 `create_chunked_extraction_job` 을 부른다. 새 잡이 생기고 **이미 끝난 청크까지 포함해 소스 전체를 처음부터 다시 추출한다.**

## 재현

1,736자 소스(6청크), 실제 LLM:

```
잡 1: 청크 0 완료 → 사실 14개, 이후 정책 삭제 → pending 으로 롤백 (청크 5개 대기)

정책 복구 후 `verinote sync`:
잡 2: 6/6 청크 → 사실 69개    ← 청크 0 포함, 전부 다시 LLM 에 보냄
잡 1: 여전히 status='pending', 청크 5개 대기       ← 아무도 안 건드림
```

LLM 호출 6회가 낭비되고, 청크 0 의 내용이 두 번 추출된다.

남은 좀비 잡이 만드는 **세 번째 추출**은 #242 에서 다룬다.

## 관련

#160 이 재추출로 인한 **중복 사실 누적**(사람이 거부한 사실의 부활 포함)을 다룬다. 이 이슈는 그 재추출이 **왜 일어나는가**를 다룬다 — 둘 다 필요하다.

## Acceptance

- `sync` 가 해당 소스의 미완 잡을 발견하면 새 잡을 만들지 않고 **재개**한다.
- 이미 `done` 인 청크를 다시 LLM 에 보내지 않는다.

기능 테스트 중 발견. #194 후속.

### LINKED ISSUE #242: fix: 좀비 잡이 다음 verinote ui 실행 때 조용히 처리되어 3중 추출이 된다
## 문제

#240 이 남긴 `pending` 좀비 잡은 사라지지 않는다. `_resume_source_extraction_jobs`(`web/app.py:782-784`)가 `pending` / `running` 잡을 **전부** 집어간다:

```python
for job in app.state.store.source_extraction_jobs():
    if job["status"] in {"pending", "running"}:
        _start_source_extraction(int(job["id"]), app.state.cfg)
```

그래서 사용자가 UI 를 켜는 것만으로, **이미 재추출이 끝난 소스의 나머지 청크가 한 번 더 추출된다.**

## 재현

#240 의 KB(잡 1 = pending 5청크, 잡 2 = 재추출 완료)에서 `verinote ui` 를 켜기만 함. **HTTP 요청 0회:**

```
BEFORE ui launch: job1=('pending', 1)   facts=83
AFTER  ui launch: job1=('done', 6)      facts=123
```

1,736자 소스 하나가 **세 번의 패스**를 거쳐 후보 사실 **123개**가 됐다 (정상은 ~69개).

## 정확한 중복이 0건이라 dedup 으로도 안 잡힌다

모델이 매번 다르게 표현하기 때문에 exact-match 중복은 0건이고, **근접 중복**으로 쌓인다. 사람이 리뷰 큐에서 같은 사실을 두 번 판정하게 된다:

```
C5 | CEO      | x2 -> ['P5', 'Person P5']
C5 | provides | x2 -> ['S5', 'Service S5']
C6 | CEO      | x2 -> ['P6', 'Person P6']
```

#160 의 dedup 만으로는 막을 수 없다는 뜻이다.

## Acceptance

- 같은 소스에 대해 미완 잡과 그것을 대체한 새 잡이 동시에 살아 있을 수 없다 (런처가 이미 대체된 잡을 되살리지 않는다).
- halt → 복구 → `sync` → `ui` 시퀀스를 태워도 청크당 LLM 호출은 **1회**다.

#240 과 함께 고쳐야 한다 — 따로 고치면 3중 추출이 2중이 될 뿐이다. #194 후속.

SeoyunL:
Closes #240
Closes #242

## The cause was not what the issue described

The issue reads as a query-condition problem — as if `sync` looked for a resumable job and looked wrong. It didn't look at all. **`cmd_sync` had no code to find an existing job**; `cli.py` called `create_chunked_extraction_job` unconditionally, every time.

The resume machinery was already complete and already correct:

- `next_pending_chunk` skips `done` chunks
- `reset_running_chunks` reclaims chunks abandoned mid-flight
- `candidate_count` accumulates across runs

A job halted mid-flight rolls back to `pending` with its finished chunks intact, ready to be picked up. Nothing ever asked for it, so the next `sync` rebuilt the job from chunk zero and paid the provider a second time for work already done.

Note that `cli.py:293` in the issue body is stale — the unconditional create was at `cli.py:472`.

## The resume predicate

`plan_source_extraction` resumes only when the job still honestly describes the work in front of us. All of these must hold:

1. **`status == 'pending'`** — `done`/`failed`/`canceled` are decided outcomes, not interrupted ones.
2. **the source's newest job** — an older job has been superseded by a later one (see #242 below).
3. **`job.artifact_id == source.artifact_id`** — artifacts are content-addressed (`UNIQUE(source_id, kind, checksum)`), so a changed body means a different artifact id.
4. **the persisted chunk texts equal what this config would produce now** — compared against `chunk_text(normalize_for_extraction(text), ...)` under the current chunk size and overlap.
5. **`job.provider == cfg.provider and job.model == cfg.model`** — resume across a provider or model switch and the job row ends up describing output it did not produce.

**Condition 4 is the load-bearing one.** It subsumes condition 3 and also catches what no identity check can see: a changed chunk size, overlap, or normalisation rule. Resume across a re-chunking and the finished chunks no longer cover the text the pending chunks assume — part of the new body falls into no chunk at all and is **silently never extracted**. The failure mode is a quietly incomplete KB with a `done` job on top of it, which is exactly the class of lie this repo exists to prevent.

A `running` job is **neither resumed nor replaced**. It may belong to a live UI worker, and resuming would have `reset_running_chunks` yank that worker's in-flight chunk back into the queue — the same chunk sent to the LLM twice. `sync` skips it and says so on stderr.

## #242: derive "superseded", do not store it

The obvious fix was to retire the dead job by writing state — `fail_extraction_job(job_id, "superseded by ...")`. That was rejected. A job replaced by a newer one is not a *failure*, and the UI renders `failed` as **"Analysis failed"**. Storing that status would put a false sentence on the user's screen about a job that did nothing wrong.

"Superseded" is instead **derived**, because it is exactly equivalent to *a newer job exists for the same source*:

```python
def is_live_extraction_job(job, latest_job_ids) -> bool:
    if job["status"] not in {"pending", "running"}:
        return False
    return int(job["id"]) == latest_job_ids.get(int(job["source_id"]))
```

Stored state can drift out of step with the rows it summarises; a derived answer cannot be wrong by construction. `sources_with_counts` already reads a source's current analysis through the same `MAX(id)` lens (`db.py:386-389`), so this is existing idiom rather than a new one. It is also entirely read-only: **`db.py` is untouched by this PR.**

## Why the scope is slightly wider than #242's letter

The launcher was the reported symptom, but the same dead `pending` row is read in two more places, and leaving them would fix the triple-extraction while keeping two broken behaviours:

- `has_running_jobs` stays `True` forever, so `sources.html:18` **polls every 2 seconds indefinitely** and the page keeps saying an analysis is in progress when nothing is processing it.
- `reanalyze_source` **409s permanently** for that source. That is not cosmetic — the re-analyse button is dead for the one source whose analysis most needs redoing.

Same predicate, same defect, one line each. All three readers now share `is_live_extraction_job`, matching the "one judgement, three enforcement points" structure `_resume_source_extraction_jobs` already documents.

## Run-scope honesty

Once a job can span runs, `outcome.candidates` is a **job total**, not this run's output. A resumed job would report "69 candidate(s)" for a run that produced 55. `_halt_extraction_job`'s docstring already insists these two scopes stay grammatically apart; enabling resume without honouring that would have introduced a fresh instance of the misreporting this work exists to remove.

`ChunkedExtractionResult` now carries `run_candidates`/`run_chunks` beside the job totals, and the sync summary states them separately:

```
sources/doc.txt: 5 candidate(s) this run (resumed job #1: 6 candidate(s) in total)
```

## A regression this PR introduced, found in review

The first version of this PR **stranded chunks that had failed in an earlier run**, and it is worth being explicit about because it was not caught by the original test plan.

`reset_running_chunks` rewinds only `'running'` chunks. `next_pending_chunk` claims only `'pending'` ones. **A `failed` chunk is neither.** A resumed job walked straight past it, reported `done`, and exited 0 — the chunk's text never reached the provider again and the facts it would have produced were lost silently.

Reproduction (six chunks, `bravo` hits a transient provider error, the policy then vanishes mid-job):

```
run 2 sent: ['charlie', 'delta', 'echo', 'foxtrot']
bravo: gone for good.  sync exit code: 0
```

**`origin/main` recovers `bravo`**, because re-syncing rebuilt the job from scratch and re-sent every chunk. Resuming removed that recovery path. This branch created the regression; it did not inherit it.

The fix is deliberately conservative — `plan_source_extraction` declines any job carrying a failed chunk, so the existing full-re-extraction behaviour stands in exactly the case that needs it. The two commits are kept unsquashed so the history shows the CRITICAL surfacing and being closed.

### The cost of that fix, and the loop it creates

Both of these are consequences of choosing the conservative fix, and neither is hidden:

1. **One failed chunk re-bills every chunk.** A six-chunk source with a single failure re-sends all six to the provider on the next `sync`. This is a deliberate trade: correctness (no silently dropped chunk) over cost.

2. **A permanently failing chunk re-extracts the whole source on every `sync`, forever.** If 5 of 6 chunks succeed every time but one always fails, `failed_chunks > 0` forces a new job on the next run, and the next, indefinitely. Because #239 is still open this loop runs at **rc 0, silently**, which makes it an unbounded supplier of the duplicate-fact accumulation tracked in #160.

This matches today's behaviour on `main` (which always re-extracts in full), so it is not a regression. But it is a real exception to this PR's headline — "sync now resumes" holds *except* when a chunk has failed, and that exception has teeth.

**Follow-up (not filed here):** retrying only the failed chunks via `retry_failed_chunks` would preserve the finished work *and* dissolve loop 2, so the cost above and the loop have a single shared remedy. It interacts with #239 and #160.

## Tests

New `tests/test_job_resume.py`, 13 tests. **The instrumentation measures the chunk text the client actually received, not which branch executed** — "the resume branch ran" is cheap to satisfy and says nothing, since a resume that re-sends chunk zero costs exactly what the bug cost.

Route-level cases build their own `TestClient` and replace the `threading` reference inside the app's namespace, making "did the launcher start a worker?" synchronous and exact rather than sleep-dependent.

- 1242 passed
- `ruff check .` clean

Every branch of the predicate is covered by a mutant with a named killer; the three conditions that had no dedicated scenario in the original plan (status, newest-job, and the *provider* half of provider/model — the model half was already locked) were found unguarded in review and now each have one.

### DIFF
--- a/tests/test_job_resume.py
+++ b/tests/test_job_resume.py
@@ -188,6 +188,43 @@ def test_sync_resumes_rolled_back_job_without_redoing_done_chunks(
     assert "sync complete: 5 candidate(s)" in out
 
 
+def test_sync_reports_a_resumed_run_that_fails_everything_as_failed_not_incomplete(
+    tmp_path, monkeypatch, capsys
+):
+    """A resumed job's OWN history must not soften this run's own total failure.
+
+    `_halted_job` leaves a job with one chunk done and zero failed (a clean
+    halt, not a chunk failure) — eligible for resume. If THIS run then fails
+    every remaining chunk, the job's cumulative `completed_chunks` is still 1
+    (from before), but this run completed none and produced nothing: the
+    "sync failed" / "sync incomplete" choice must be judged on what THIS run
+    did, not smoothed over by a resumed job's earlier, unrelated success.
+    """
+    job_id = _halted_job(tmp_path, monkeypatch)
+
+    class _AlwaysFailingClient:
+        name = "fake"
+
+        def extract_facts(self, *, source_text: str, schema_hint: str = ""):
+            raise LLMError("provider down")
+
+    monkeypatch.setattr("verinote.llm.get_client", lambda cfg: _AlwaysFailingClient())
+
+    assert cli.main(["sync"]) == 1
+
+    jobs = _jobs(tmp_path)
+    assert [int(job["id"]) for job in jobs] == [job_id]  # still resumed, not replaced
+    assert int(jobs[0]["completed_chunks"]) == 1  # unchanged: this run finished none
+
+    out, err = capsys.readouterr()
+    # This run's own contribution is honestly zero, not the job's stale total.
+    assert "0 candidate(s) this run" in out
+    assert f"resumed job #{job_id}: 1 candidate(s) in total" in out
+    # The judged-on-this-run-alone outcome: every chunk THIS run touched failed.
+    assert "sync failed" in err
+    assert "sync incomplete" not in err
+
+
 # --- the chunk a resume can never reach -------------------------------------
 
 
@@ -459,6 +496,27 @@ def test_sync_leaves_a_running_job_alone(tmp_path, monkeypatch, capsys):
     store.close()
     err = capsys.readouterr().err
     assert f"extraction job #{job_id} is already running" in err
+
+
+def test_sync_skips_a_job_claimed_between_plan_and_process(tmp_path, monkeypatch, capsys):
+    """Plan sees `pending`; another worker wins the claim first; sync skips cleanly (#240).
+
+    The plan-time `busy_job_id` branch only catches a job already `running` when the
+    plan is drawn. A job still `pending` at plan time can be claimed by a competing
+    worker in the window before `process_extraction_job` runs; the raised
+    `ExtractionJobBusyError` must be the same skip, never a crash or a failure.
+    """
+    job_id = _halted_job(tmp_path, monkeypatch)
+    monkeypatch.setattr(Store, "claim_pending_extraction_job", lambda self, jid: False)
+    monkeypatch.setattr("verinote.llm.get_client", lambda cfg: _RefusingClient())
+
+    assert cli.main(["sync"]) == 0  # skipped cleanly: no crash, no non-zero exit
+
+    err = capsys.readouterr().err
+    assert f"extraction job #{job_id} is already running" in err
+    jobs = _jobs(tmp_path)
+    assert [int(job["id"]) for job in jobs] == [job_id]  # no replacement job created
+    assert jobs[0]["status"] == "pending"  # left untouched for its real owner
 
 
 # --- D, F, G: the superseded job is dead, and all three readers agree --------
--- a/tests/test_job_resume.py
+++ b/tests/test_job_resume.py
@@ -23,7 +23,7 @@ import pytest
 
 import verinote.cli as cli
 from verinote.engine import DEFAULT_POLICY
-from verinote.llm.base import ExtractedFact
+from verinote.llm.base import ExtractedFact, LLMError
 from verinote.pipeline import create_chunked_extraction_job
 from verinote.pipeline.policy_state import POLICY_RELPATH
 from verinote.store import Store
@@ -54,10 +54,17 @@ class _RecordingClient:
 
     name = "fake"
 
-    def __init__(self, *, delete_policy_on_call: int | None = None, policy_path=None):
+    def __init__(
+        self,
+        *,
+        delete_policy_on_call: int | None = None,
+        policy_path=None,
+        fail_on_call: int | None = None,
+    ):
         self.seen: list[str] = []
         self._delete_on = delete_policy_on_call
         self._policy_path = policy_path
+        self._fail_on = fail_on_call
 
     @property
     def markers(self) -> list[str]:
@@ -65,6 +72,8 @@ class _RecordingClient:
 
     def extract_facts(self, *, source_text: str, schema_hint: str = ""):
         self.seen.append(source_text)
+        if self._fail_on is not None and len(self.seen) == self._fail_on:
+            raise LLMError("provider down")  # transient: the chunk goes `failed`
         if self._delete_on is not None and len(self.seen) == self._delete_on:
             self._policy_path.unlink()
         return [ExtractedFact(source_text.split()[0], "seen_in", "source", 0.9)]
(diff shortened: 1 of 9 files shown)


justinjoy (changes requested):
차단 이슈가 있습니다.

`cmd_sync()`가 `plan_source_extraction()`으로 pending job 재개를 결정한 뒤, `process_extraction_job()`는 별도의 소유권 획득 없이 `reset_running_chunks(job_id)`를 먼저 호출하고 이어서 job을 running으로 표시합니다. 이 사이에 다른 `verinote sync` 프로세스나 UI startup worker가 같은 rolled-back job을 먼저 잡으면, 뒤늦게 들어온 worker가 앞선 worker의 in-flight chunk를 다시 `pending`으로 되돌릴 수 있습니다. 그 결과 같은 chunk가 LLM에 중복 전송되거나, 더 늦은 write가 앞선 진행 상태를 덮을 수 있습니다.

현재 테스트는 planning 전에 이미 `running` 상태인 job만 다루고 있어, `pending`으로 관찰한 뒤 처리 시작 전 다른 worker가 먼저 소유하는 plan-to-process race를 검증하지 못합니다.

재개 경로에는 atomic ownership claim이 필요합니다. 예를 들어 `pending -> running` 전이를 `WHERE id = ? AND status = "pending"` 같은 조건으로 성공 여부를 확인한 뒤, 성공한 worker만 `reset_running_chunks()`와 chunk 처리로 들어가도록 해 주세요. 실패한 worker는 busy/skip으로 빠져야 합니다.

PR #320, opened 2026-07-20
fix(#293): make a blank VERINOTE_BASE_URL mean the same thing to every provider

### LINKED ISSUE #293: fix(config): an empty VERINOTE_BASE_URL means different things to different adapters
## 배경

PR #254 (README) 리뷰 중 발견됐다. README 주장을 반증하지는 않아 그 PR 범위 밖으로 뒀다.

## 문제

`config.py` 의 `_pick()` 이 같은 함수 안에서 빈 값을 **두 가지로** 취급한다:

```python
def _pick(env: str, saved: str | None, default: str | None) -> str | None:
    value = os.environ.get(env)
    if value is not None:      # <-- 빈 문자열이 통과한다
        return value
    if saved:                  # <-- 여기는 truthy 검사
        return saved
    ...
```

env 만 falsy 폴백에서 빠진다. 그 결과 `export VERINOTE_BASE_URL=` (빈 값) 하나로 provider 마다 다른 일이 벌어진다:

| 어댑터 | `base_url=""` 일 때 |
|---|---|
| Ollama | `cfg.base_url or "http://localhost:11434"` → falsy 라 **조용히 localhost 폴백** |
| Anthropic / OpenAI | `base_url=""` 를 그대로 SDK 에 전달 → **깨짐** |

## 왜 고쳐야 하나

1. 오타 한 번(`export VERINOTE_BASE_URL=`)으로 밟히고, 증상이 provider 마다 달라 원인 추적이 어렵다.
2. 이 프로젝트의 "조용히 틀리는 것이 최악" 원칙에 걸린다. `VERINOTE_BASE_URL=""` 는 **사용자 의도가 불명확한 입력**인데, 지금은 provider 에 따라 조용히 기본값으로 돌아가거나 SDK 레벨에서 엉뚱한 에러를 낸다.

## 할 일

`_pick` 이 빈 문자열을 **미설정으로 정규화**하든, 아니면 **시끄럽게 거부**하든 한 가지로 정한다. 어느 쪽이 맞는지는 설정 계층 전반의 결정이다 (`VERINOTE_BASE_URL` 만의 문제가 아니라 `_pick` 을 쓰는 모든 설정에 해당).

## 참고

- PR #254 — https://github.com/semantic-reasoning/verinote/pull/254

SeoyunL:
`export VERINOTE_BASE_URL=` currently means a different thing to every provider. This makes a blank value mean *unset*, everywhere.

Closes #293

## Commit title correction

The title of `48e9e40` says `VERINOTE_*`, which overclaims. The change applies to the three settings that go through `_pick`: **`VERINOTE_PROVIDER`, `VERINOTE_MODEL`, and `VERINOTE_BASE_URL`**. The other environment readers are untouched. Amending was off the table once the branch had review history, so the correction lives here.

## What was broken

The four adapters each did something different with `""`:

| adapter | behaviour with `base_url=""` |
|---|---|
| `openai_adapter.py` | passes `""` straight to the SDK |
| `anthropic_adapter.py` | passes `""` straight to the SDK |
| `ollama_adapter.py` | `""` is falsy, so it silently uses localhost |
| `claude_cli_adapter.py` | never reads `base_url` at all |

Measured against the installed anthropic SDK: `base_url=""` fails at request time with `APIConnectionError: Connection error.` — an opaque error that names no cause. `base_url=None` resolves to `https://api.anthropic.com` **and honours the SDK's own `ANTHROPIC_BASE_URL`**. So an empty string does not just fail; it disables the SDK's own environment handling.

The cause was one place. `_pick` tested the environment with `is not None`, so `""` passed through. The module's other five readers already treat blank as unset (`local_root`, `active_root`, `_llm_timeout_seconds`, `_pick_int`, `_pick_bool`), and both write paths already normalise (`save_settings`, `web/app.py`). `_pick` was the lone exception.

## What changed, and why it took three commits

Review found the second half of the bug twice. That is what the three commits are.

1. **`48e9e40`** — normalise the environment source. **But the `saved` branch was still `if saved:`**, so the fix covered half the read path and the same bug walked back in through the Settings UI: `save_settings(base_url="   ")` produced an ollama request to `'   /api/chat'`.

2. **`75b4f2d`** — fold both sources into one loop and trim the value that is used. Judging on the trimmed text while returning the raw one is its own bug: `"  https://x/v1  "` builds `'...v1  /api/chat'`.

3. **`d8f9671`** — guard the provider and model halves. Narrowing the trim to `env == "VERINOTE_BASE_URL"` with a single line left **all ~30 tests green**: `_pick` serves three settings, but every mutant so far had aimed at `base_url`.

`_pick` now reads:

```python
for candidate in (os.environ.get(env), saved):
    if isinstance(candidate, str):
        candidate = candidate.strip()
    if candidate:
        return candidate
return default
```

No adapter source changed. The `isinstance` check preserves the existing pass-through for a non-string value in a hand-edited `config.json`.

## Backward compatibility

No working configuration breaks. The only users relying on a blank value today are on ollama, and their behaviour is unchanged (localhost either way). anthropic and openai users are broken today and are fixed here.

`VERINOTE_PROVIDER=""` now falls back instead of raising `unknown VERINOTE_PROVIDER=''`. That is an improvement, not a loss of safety: a typo does not produce a blank value, so `opnai` still fails loudly.

## Known remaining inconsistency

The module still has two rules after this branch. `_pick` falls back to the saved file on a blank env value; **`_pick_int` and `_pick_bool` skip the saved file and go straight to the built-in default.** Measured: saved `chunk_chars=450` with `VERINOTE_EXTRACTION_CHUNK_CHARS=""` yields **300**, while the same shape on `VERINOTE_MODEL=""` yields the saved value. The new module docstring spells this split out, so it is documented rather than hidden — but it is not unified. Left as follow-up.

## Verification

`1262 passed`, 0 failed. `ruff check .` clean.

Seven mutants, all red (unmutated copy of the same tree: 73 passed):

| mutant | result |
|---|---|
| revert `_pick` entirely | 23 failed |
| normalise but never fall back | 9 failed |
| judge on trimmed text, return raw | 7 failed |
| revert the `saved` branch only | 5 failed |
| restrict the trim to `VERINOTE_BASE_URL` | 3 failed |
| `openai_adapter` regains `or ""` | 2 failed |
| ollama default becomes `127.0.0.1` | 2 failed |

The adapter-layer guards assert `recorded["base_url"] is None` rather than a falsy check — `""` is falsy too, so a truthiness assertion would pass against the very bug. The ollama guards capture the **requested URL**, not the adapter attribute.

Three reverse guards confirm the normalisation is not over-constrained, so self-hosted endpoints still work: a custom endpoint reaches openai/anthropic verbatim and ollama as `https://llm.internal/v1/api/chat`, trailing-slash stripping survives, and a `config.json`-only endpoint still reaches the adapter with no env set at all.

## Follow-ups (not filed here)

- Non-string config values belong in `read_settings` type validation — that is the boundary where untrusted JSON arrives, and it already warns-and-ignores a non-dict. Fixing it in `_pick` would miss `_pick_int`/`_pick_bool` and the direct `saved.get()` call sites.
- `VERINOTE_API_KEY=""` bypasses `_pick` and is read directly.
- The `_pick_int`/`_pick_bool` saved-skipping described above.

Deliberately **not** doing base_url format validation: there is no clean boundary for a valid endpoint (self-hosted, ports, sockets), and with blank values gone the remaining failures already point at their cause. It would be exactly the over-constraint the reverse guards exist to prevent.

### DIFF
--- a/tests/test_cloud_adapters.py
+++ b/tests/test_cloud_adapters.py
@@ -291,3 +291,27 @@ def test_settings_file_base_url_reaches_cloud_client(tmp_path, monkeypatch, prov
     adapter_cls(Config.for_root(tmp_path)).extract_facts(source_text="x")
 
     assert recorded["base_url"] == "https://llm.internal/v1"
+
+
+@pytest.mark.parametrize("provider", sorted(_CLOUD_ADAPTERS))
+def test_whitespace_only_saved_base_url_reaches_cloud_client_as_none(tmp_path, monkeypatch, provider):
+    record, adapter_cls = _CLOUD_ADAPTERS[provider]
+    recorded = record(monkeypatch)
+    monkeypatch.delenv("VERINOTE_BASE_URL", raising=False)
+    save_settings(tmp_path, provider=provider, model="m", base_url="   ")
+
+    adapter_cls(Config.for_root(tmp_path)).extract_facts(source_text="x")
+
+    assert recorded["base_url"] is None
+
+
+@pytest.mark.parametrize("provider", sorted(_CLOUD_ADAPTERS))
+def test_padded_base_url_reaches_cloud_client_trimmed(tmp_path, monkeypatch, provider):
+    record, adapter_cls = _CLOUD_ADAPTERS[provider]
+    recorded = record(monkeypatch)
+    monkeypatch.setenv("VERINOTE_PROVIDER", provider)
+    monkeypatch.setenv("VERINOTE_BASE_URL", "  https://llm.internal/v1  ")
+
+    adapter_cls(Config.for_root(tmp_path)).extract_facts(source_text="x")
+
+    assert recorded["base_url"] == "https://llm.internal/v1"
--- a/tests/test_cloud_adapters.py
+++ b/tests/test_cloud_adapters.py
@@ -4,7 +4,7 @@ from types import SimpleNamespace
 
 import pytest
 
-from verinote.config import Config
+from verinote.config import Config, save_settings
 from verinote.llm.anthropic_adapter import AnthropicAdapter
 from verinote.llm.base import LLMError
 from verinote.llm.openai_adapter import OpenAIAdapter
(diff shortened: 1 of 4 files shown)


justinjoy (changes requested):
차단 이슈가 있습니다.

`verinote/config.py`의 module docstring이 새로 “numeric and boolean settings drop straight to the built-in default without consulting the saved file”이라고 설명하지만, 실제 구현은 그렇지 않습니다. `Config.for_root()`는 `VERINOTE_EXTRACTION_CHUNK_CHARS`, `VERINOTE_EXTRACTION_CHUNK_OVERLAP_CHARS`, `VERINOTE_EXTRACTION_MAX_FACTS_PER_CHUNK`, `VERINOTE_AUTO_ACCEPT_RECOMMENDATIONS`에 대해 env가 없으면 `config.json`의 saved 값을 `_pick_int()` / `_pick_bool()`에 넘기고, 기존 `test_extraction_settings_round_trip_and_env_override`도 이 동작을 보장합니다.

즉 이 PR을 병합하면 설정 precedence를 설명하는 최상단 문서가 실제 런타임 계약과 반대로 기록됩니다. blank normalization 설명은 유지하되, numeric/boolean 설정도 env -> saved -> default precedence를 따른다는 실제 동작에 맞게 문장을 수정해 주세요.

기능 검증으로 `tests/test_config.py tests/test_cloud_adapters.py tests/test_ollama_adapter.py`는 로컬에서 통과했습니다.

justinjoy (approved):
변경요청 사항이 반영되었습니다. config module docstring이 provider/model/base URL의 blank fallthrough, numeric parser fallback, boolean parser 해석, API key env-only 동작을 실제 구현과 일치하게 설명합니다. read_settings 경고 문구도 저장 가능한 runtime settings 범위에 맞게 갱신되었습니다. 로컬 및 CI 검증이 통과했습니다.

